PR #4, opened 2020-06-04
Fix CPU test failures

adayton1:
This gets all the tests passing for the CPU only build.

### DIFF
--- a/test/TestScan.cpp
+++ b/test/TestScan.cpp
@@ -22,8 +22,9 @@ using int_ptr = chai::ManagedArray<int>;
 
 CUDA_TEST(Scan, test_scan_offset) {
 #if defined(__CUDACC__)
-   care::initialize_pool("PINNED","PINNED_POOL",chai::PINNED,128*1024*1024,128*1024*1024,true);
-   care::initialize_pool("DEVICE","DEVICE_POOL",chai::GPU,128*1024*1024,128*1024*1024,true);
+   int poolSize = 128*1024*1024; // 128 MB
+   care::initialize_pool("PINNED", "PINNED_POOL", chai::PINNED, poolSize, poolSize ,true);
+   care::initialize_pool("DEVICE", "DEVICE_POOL", chai::GPU, poolSize, poolSize, true);
 #endif
 
    const int starting_offset = 17;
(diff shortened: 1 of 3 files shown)


test/TestScan.cpp
```diff
@@ -21,8 +21,11 @@
 using int_ptr = chai::ManagedArray<int>;
 
 CUDA_TEST(Scan, test_scan_offset) {
+#if defined(__CUDACC__)
    care::initialize_pool("PINNED","PINNED_POOL",chai::PINNED,128*1024*1024,128*1024*1024,true);
    care::initialize_pool("DEVICE","DEVICE_POOL",chai::GPU,128*1024*1024,128*1024*1024,true);
```
dtaller, line 26: Could you add a small comment on where the magic number 128*1024*1024,128*1024*1024 comes from?

test/TestSortFuser.cxx
```diff
@@ -18,8 +18,11 @@ using namespace care;
 using int_ptr = host_device_ptr<int>;
 
 CUDA_TEST(TestPacker, testFuseSort) {
+#if defined(__CUDACC__)
    care::initialize_pool("PINNED","PINNED_pool",chai::PINNED,128*1024*1024,128*1024*1024,true);
    care::initialize_pool("DEVICE","DEVICE_pool",chai::GPU,128*1024*1024,128*1024*1024,true);
```
dtaller, line 23: A comment on the value 128*1024*1024,128*1024*1024 would be nice here also.

test/TestSortFuser.cxx
```diff
@@ -18,16 +18,19 @@ using namespace care;
 using int_ptr = host_device_ptr<int>;
 
 CUDA_TEST(TestPacker, testFuseSort) {
-   care::initialize_pool("PINNED","PINNED_pool",chai::PINNED,128*1024*1024,128*1024*1024,true);
-   care::initialize_pool("DEVICE","DEVICE_pool",chai::GPU,128*1024*1024,128*1024*1024,true);
+#if defined(__CUDACC__)
```
robinson96, line 21: does this actually fail if __CUDACC__ is disabled? Do we want to enable it with CHAI_GPU_SIM_MODE ?

PR #5, opened 2020-06-04
Feature/taller1/arrayutiltests

dtaller:
Tests for array_utils, and some related documentation and bugfixes

### DIFF
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -570,6 +570,10 @@ inline int BinarySearch(const T *map, const int start,
       }
       // the upper option is within the range of the map index set
       if (khi < start + mapSize) {
+         while (map[khi] == num && khi < start + mapSize) { // Note: fix for last test in TEST(array_utils, binarysearch)
+            ++khi;
+         }
+
          // the upper option bounds num
          if (map[khi] > num) {
             return khi;
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -1309,6 +1309,7 @@ inline T ArraySum(care::host_device_ptr<const T> arr, int n, T initVal)  {
  * Function  : ArraySumSubset
  * Author(s) : Peter Robinson
  * Purpose   : Returns the sum of values in arr at indices in subset.
+ * Note      : length n refers to length of subset, not array
  * ************************************************************************/
 template <typename T, typename ReduceType, typename Exec>
 inline T ArraySumSubset(care::host_device_ptr<const T> arr, care::host_device_ptr<int const> subset, int n, T initVal) {
@@ -1370,7 +1371,7 @@ inline int FindIndexGT(care::host_device_ptr<const T> arr, int n, T limit) {
    } LOOP_REDUCE_END
    return maxLoc.getLoc();
 
-   /* above is supposed to be equivalent to below sequenential code:
+   /* above is supposed to be equivalent to below sequenential code, but it isn't.
    int i ;
    for (i=0 ; i<n ; ++i) {
       if (arr[i] > limit) {
@@ -1378,6 +1379,10 @@ inline int FindIndexGT(care::host_device_ptr<const T> arr, int n, T limit) {
       }
    }
    return -1 ;
+
+   Note: It is  NOT equivalent to that code. The provided code finds the index which is the most above
+   the limit. for example, for an array [0, 1, 2, 3, 4] with limit 2, the sequential code returns 3 but the
+   LOOP_REDUCE implementation return 4.
    */
 }
 
@@ -1389,7 +1394,7 @@ inline int FindIndexGT(care::host_device_ptr<const T> arr, int n, T limit) {
  * ************************************************************************/
 template <typename T, typename Exec>
 inline int FindIndexMax(care::host_device_ptr<const T> arr, int n) {
-   RAJAReduceMaxLoc<T> maxLoc { -FLT_MAX, -1 };
+   RAJAReduceMaxLoc<T> maxLoc { std::numeric_limits<T>::lowest(), -1 };
    LOOP_REDUCE(i, 0, n) {
       maxLoc.maxloc(arr[i], i);
    } LOOP_REDUCE_END
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -1443,12 +1443,12 @@ inline void ArrayCopy(RAJA::seq_exec,
 /************************************************************************
  * Function  : ArrayDup
  * Author(s) : Peter Robinson
- * Purpose   : Duplicates a ManagedArray. If input is null or len 0, 
+ * Purpose   : Duplicates a ManagedArray. If input is nullptr, 
  *             returns a nullptr (no allocation occurs).
  * ************************************************************************/
 template <typename T, typename Exec>
 inline care::host_device_ptr<T> ArrayDup(care::host_device_ptr<const T> from, int len) {
-   if (from == nullptr || len <= 0) { // don't make a new array for null input
+   if (from == nullptr) { // don't make a new array for null input
      return nullptr;
    } else {
      care::host_device_ptr<T> newArray(len,"ArrayDup newArray");
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -1229,10 +1229,10 @@ int ArrayMinMax(care::host_device_ptr<T> arr, care::host_device_ptr<int> mask, i
 template <typename Exec=RAJAExec>
 int ArrayMinMax(care::host_device_ptr<const globalID> arr, care::host_device_ptr<int const> mask, int n, double *outMin, double *outMax);
 
-//template <typename Exec>
-//int ArrayMinMax(care::host_device_ptr<const globalID> arr, care::host_device_ptr<int const> mask, int n, double *outMin, double *outMax) {
-//   return ArrayMinMax<globalID, GIDTYPE, Exec>(arr, mask, n, outMin, outMax);
-//}
+template <typename Exec>
+int ArrayMinMax(care::host_device_ptr<const globalID> arr, care::host_device_ptr<int const> mask, int n, double *outMin, double *outMax) {
+   return ArrayMinMax<globalID, GIDTYPE, Exec>(arr, mask, n, outMin, outMax);
+}
 
 #endif // CARE_HAVE_LLNL_GLOBALID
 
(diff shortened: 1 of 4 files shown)


src/care/array_utils.h
```diff
@@ -519,7 +549,8 @@ inline int BinarySearch(const T *map, const int start,
       return -1 ;
    }
 #ifdef CARE_DEBUG
-   checkSorted(&(map[start]), mapSize, "BinarySearch", "map") ;
+   const bool allowDuplicates = true;
```
robinson96, line 552: the comment change on L 478 and this line of code appear to conflict with each other.

src/care/array_utils.h
```diff
@@ -1157,6 +1219,11 @@ int ArrayMinMax(care::host_device_ptr<const T> arr, care::host_device_ptr<int co
    return (int) result;
 }
 
+template <typename T, typename ReducerType, typename RAJAExec>
+int ArrayMinMax(care::host_device_ptr<T> arr, care::host_device_ptr<int> mask, int n, double *outMin, double *outMax) {
```
robinson96, line 1223: inline

src/care/array_utils.h
```diff
@@ -1309,14 +1386,18 @@ inline int FindIndexGT(care::host_device_ptr<const T> arr, int n, T limit) {
...
+   /* above is supposed to be equivalent to below sequenential code, but it isn't.
    int i ;
    for (i=0 ; i<n ; ++i) {
       if (arr[i] > limit) {
          return i ;
       }
    }
    return -1 ;
+
+   Note: It is  NOT equivalent to that code. The provided code finds the index which is the most above
```
robinson96, line 1398: Then this is a bug that we should fix. 
```
   RAJAReduceMin<int> minIndexAboveLimit {n};
   LOOP_REDUCE(i, 0, n) {
     if ( arr[i] > limit) {
          minIndexAboveLimit.min(i);
      }
   } LOOP_REDUCE_END
   return (int) minIndexAboveLimit;
```

test/TestArrayUtils.cpp
```diff
@@ -0,0 +1,1214 @@
...
+  // test init val -1
+  initVal = -1;
+  result = care_utils::ArrayMin<int>(a, 7, initVal, 0);
+  EXPECT_EQ(result, -1);
+}
+
+GPU_TEST(array_utils, max_empty)
+{
+  care::host_device_ptr<int> a;
+  // ArrayMin has a value for start index but ArrayMax does not. TODO: use slicing or pointer arithmatic instead.
```
robinson96, line 292: can you add a github ticket for this?

dtaller: Added two tickets:
1) For making the start index issue consistent, see https://rzlc.llnl.gov/jira/browse/ALE-2364 
2) More generally, for initial values when doing max, we sometimes use user-provided initial values, sometimes explicitly assume a double type, and sometimes (host only, standard library not availible on the device) use std::numeric_limits<T> . We should make this all consistent, see https://rzlc.llnl.gov/jira/browse/ALE-2365

src/care/array_utils.h
dtaller: I think this is right. This is in BinarySearch. If I understand correctly, BinarySearch does allow duplicates (it will find an element with the desired value, and if others with the same value exist nearby they will be adjacent, either before or after)

However, IntersectArrays does not allow duplicates, because we have to return the indicies of all the matches and that would be with duplicates.

src/care/host_device_ptr.h
```diff
@@ -102,12 +102,14 @@ namespace care {
 
       ///
       /// @author Peter Robinson
-      ///
+      /// @note   cannot make this CARE_HOST_DEVICE due to use of standard library.
+      //          The test file TestArrayUtils.cpp gives a warning about calling a
+      //          __host__ function from a __host__ __device__ function when compiled on rzansel
```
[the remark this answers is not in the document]
dtaller: Changed comment

src/care/array_utils.h
dtaller: Changed, thanks

src/care/array_utils.h
```diff
@@ -1309,14 +1386,18 @@ inline int FindIndexGT(care::host_device_ptr<const T> arr, int n, T limit) {
    } LOOP_REDUCE_END
    return maxLoc.getLoc();
 
-   /* above is supposed to be equivalent to below sequenential code:
+   /* above is supposed to be equivalent to below sequenential code, but it isn't.
```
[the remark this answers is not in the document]
dtaller: Changed

src/care/array_utils.h
dtaller: Done, made the change

dtaller:
> Overall really good work, Danny. Glad you uncovered some issues (also surprised the bugs you found are not causing issues in our integration tests. It might be worth hunting for uses in our codebase to see if we are just getting lucky, and see if you can come up with an integration test in ale3d that exposes these underlying bugs.

Thanks! I'll take a look

src/care/array_utils.h
```diff
@@ -1177,29 +1244,32 @@ int ArrayMinMax(care::host_device_ptr<const globalID> arr, care::host_device_ptr
  *             Otherwise, returns 1.
  *             care::local_ptr API to support calls from within RAJA contexts.
  * ************************************************************************/
-template <typename T, typename ReducerType>
-inline int ArrayMinMax(care::local_ptr<const T> arr, care::local_ptr<int const> mask, int n, double *outMin, double *outMax) {
+template <typename T>
+CARE_HOST_DEVICE inline int ArrayMinMax(care::local_ptr<const T> arr, care::local_ptr<int const> mask, int n, double *outMin, double *outMax) {
    bool result = false;
-   ReducerType min, max;
+   // a previous implementation had min and max as a templated type and then used std::numeric_limits<T>::lowest() and
```
adayton1, line 1250: Add an issue for this.

src/care/array_utils.h
```diff
@@ -1490,7 +1582,9 @@ int FindIndexMinSubset(care::host_device_ptr<const T> arr, care::host_device_ptr
 // @param subset     : The subset of arr to find the min value for.
 // @param lenset     : Length of subset.
 // @param thresholds : If thresholds is not nullptr, only look at indices where
-//                     thresholds is above cutoff. length >= max(subset[0:lenset])
+//                     thresholds is above cutoff. length >= max(subset[0:lenset]).
```
adayton1, line 1585: Thanks for all these clarifications!

src/care/host_device_ptr.h
```diff
@@ -102,12 +102,14 @@ namespace care {
 
       ///
       /// @author Peter Robinson
-      ///
+      /// @note   cannot make this CARE_HOST_DEVICE due to use of standard library.
```
adayton1, line 105: I'm surprised by this.

test/TestArrayUtils.cpp
```diff
@@ -0,0 +1,1214 @@
...
+
+// care headers
+#include "care/array.h"
+#include "care/array_utils.h"
+#include "care/care.h"
+
+// Array Fill Tests
+TEST(array_utils, fill_empty)
+{
+   care::host_ptr<int> a;
```
adayton1, line 24: Which ArrayFill method is being called with a host_ptr? Does it call the local_ptr method? I'm thinking we should make these host_device_ptrs.

test/TestArrayUtils.cpp
```diff
@@ -0,0 +1,1214 @@
...
+
+// Adapted from CHAI
+#define GPU_TEST(X, Y) \
+   static void gpu_test_##X##Y(); \
+   TEST(X, gpu_test_##Y) { gpu_test_##X##Y(); } \
+   static void gpu_test_##X##Y()
+
+GPU_TEST(array_utils, fill_empty) {
+   care::host_device_ptr<int> a;
+   care_utils::ArrayFill<int>(a, 0, -12); // hopefully nothing explodes
```
adayton1, line 189: Could do EXPECT_EQ(a, nullptr)

test/TestArrayUtils.cpp
```diff
@@ -0,0 +1,1214 @@
...
+   care::host_device_ptr<int> b(tempb, 1, "fillhostdev");
+
+   care_utils::ArrayFill<int>(b, 1, -12);
+
+   EXPECT_EQ(b.pick(0), -12);
+}
+
+GPU_TEST(array_utils, fill_three) {
+   int tempb[3] = {1, 2, 3};
+   // if you attempt to initialize this as b(tempb) without the size, the test fails. Maybe that constructor should be
```
adayton1, line 206: Add an issue for this.

test/TestArrayUtils.cpp
dtaller: I made github tickets also. Those were jira tickets, I was confused

src/care/array_utils.h
dtaller: I made github tickets. Previously I got confused and made jira tickets instead

src/care/array_utils.h
dtaller: No problem. It came up when making tests. Hopefully I save future users time

test/TestArrayUtils.cpp
dtaller: Added github issue

dtaller:
Wait, don't merge yet. Two tests have different results when combined in release vs debug mode, I'll investigate.

adayton1:
Any progress, Danny?

dtaller:
> Any progress, Danny?

Oh sorry, I didn't resolve it yesterday, and I got distracted with some other tasks today. I'll get back to it soon. Yesterday I found that if I did the same test twice in a row, same exact inputs, it gave different results. There must be some kind of memory issue or overwrite that I overlooked. One array had some invalid values in it even though it should not have been modified, but I have not put my finger on it exactly.

adayton1:
OK no hurry. I just wanted to make sure I hadn't missed something.

adayton1:
Also, if you want to create issues for those tests, I can still merge this in and one of us can look at them later.

dtaller:
> Also, if you want to create issues for those tests, I can still merge this in and one of us can look at them later.

I'll probably finish some cnfd stuff and then take another look. If I am still lost, I might take you up on that offer. Another pair of eyes would be useful if I still can't find it after another try.

dtaller:
> Also, if you want to create issues for those tests, I can still merge this in and one of us can look at them later.

Hi Alan. I managed to make one failure go away, but there is still one test that has inconsistent results between the rzansel debug and relwithdebinfo runs. You can merge this in, I'll make an issue to solve in the future.

So, in the future, if anyone adds any functions to array_utils, we can request that they add tests in the corresponding test file.

PR #10, opened 2020-06-05
Adds a couple of examples

adayton1:
This pull request adds examples of using a STREAM_LOOP and a SCAN_LOOP.

### DIFF
--- a/examples/ScanLoop.cpp
+++ b/examples/ScanLoop.cpp
@@ -53,7 +53,9 @@ int main(int, char**) {
    // Set the size
    int size2 = 1000;
 
-   // Allocate memory on the host and device using a CARE wrapper around a CHAI container
+   // Allocate memory on the host and device using a CARE wrapper around a CHAI container.
+   // This object can be used on both the host and device, so it can be used in a
+   // LOOP_STREAM. Raw arrays should not be used inside a LOOP_STREAM.
    care::host_device_ptr<int> data2(size2);
 
    // Set a conditional
--- a/examples/ScanLoop.cpp
+++ b/examples/ScanLoop.cpp
@@ -1,0 +1,58 @@
+// This macro needs to be defined before including care/care.h,
+// which allows you to port a file at a time. Without this define,
+// LOOP_STREAM will run on the CPU. With this define, LOOP_STREAM
+// will run on the GPU only if you built with CUDA enabled. Otherwise,
+// if CUDA is disabled, it will default back to running on the host.
+#define GPU_ACTIVE
+
+// CARE library header
+#include "care/care.h"
+#include "care/scan.h"
+
+// Std library headers
+#include <iostream>
+#include <vector>
+
+int main(int, char**) {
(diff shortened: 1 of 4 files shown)


examples/ScanLoop.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+int main(int, char**) {
+   // Normal code
+   int size = 1000;
+   std::vector<int> data;
+
+   int threshold = 420;
+
+   for (int i = 0; i < size; ++i) {
+      int value = i * i;
```
dtaller, line 31: Since this is under examples, I would be very generous with comments. Make comments as if you are explaining to code to someone who has never seen RAJA or care .

examples/ScanLoop.cpp
```diff
@@ -0,0 +1,65 @@
...
+      if (value < threshold) { 
+         data.push_back(value);
+      }
+   }
+
+   std::cout << "CPU count: " << data.size() << std::endl;
+
+   // Ported code
+   int size2 = 1000;
+   care::host_device_ptr<int> data2(size2);
```
dtaller, line 42: Comment on why you are creating a host_device pointer, which can has data on both the GPU and CPU. Without this, the SCAN_LOOP would be invalid, you can't use a regular c array etc etc

examples/ScanLoop.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+   int count = 0;
+   int threshold2 = 420;
+
+   SCAN_LOOP(i, 0, size2, pos, count, i * i < threshold2) {
+      data2[pos] = i * i;
```
dtaller, line 48: Comment on each argument of the SCAN LOOP, what the purpose is. Since this is under examples, explain even basic things thoroughly

examples/StreamLoop.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+   int max = *std::max_element(data.begin(), data.end());
+   std::cout << "CPU max: " << max << std::endl;
+
+   // Ported code
+   int size2 = 1000;
+   care::host_device_ptr<int> data2(size2);
+
+   LOOP_STREAM(i, 0, size2) {
+      data2[i] = i * i;
```
dtaller, line 42: Likewise in this file and elsewhere, be very generous with comments in examples

examples/ScanLoop.cpp
adayton1: Done.

examples/ScanLoop.cpp
adayton1: Done.

examples/ScanLoop.cpp
adayton1: Provided the equivalent code so they can understand what the macro is doing.

examples/StreamLoop.cpp
adayton1: Done.

PR #14, opened 2020-06-23
Add BLT submodule

adayton1:
If BLT_SOURCE_DIR is set, use the external BLT indicated. Otherwise, use the git submodule.


adayton1:
Host codes are using BLT v0.3.0.

PR #17, opened 2020-07-06
Add git submodules for dependencies

adayton1:
This pull request adds git submodules for dependencies.


CMakeLists.txt
```diff
@@ -18,19 +18,77 @@ option(ENABLE_MPI "Build MPI support" OFF)
...
-set(camp_DIR "" CACHE PATH "Path to external campConfig.cmake or camp-config.cmake")
+
+# TODO: When RAJA is updated past v0.11.0, camp_DIR should point to the
+#       directory containing campConfig.cmake or camp-config.cmake.
+#       Update the variable documentation to reflect that change.
+set(camp_DIR "" CACHE PATH "Path to external CAMP (top level of the install or source directory)")
+
+set(umpire_DIR "" CACHE PATH "Path to external umpireConfig.cmake or umpire-config.cmake")
+set(raja_DIR "" CACHE PATH "Path to external rajaConfig.cmake or raja-config.cmake")
+set(chai_DIR "" CACHE PATH "Path to external chaiConfig.cmake or chai-config.cmake")
```
dtaller, line 54: why are these lower case (chai_DIR)? Wouldn't uppercase be more consistent with other options (CHAI_DIR) ?

cmake/SetupLibraries.cmake
```diff
@@ -28,72 +28,6 @@ else()
     set(CARE_HAVE_BASIL "0" CACHE STRING "")
 endif()
 
-################################
-# UMPIRE
-################################
-if (UMPIRE_DIR)
-    include(cmake/libraries/Findumpire.cmake)
```
dtaller, line 35: If the user wants to have an external umpire (like if we are compiling ale3d), will things still work? Or does it only work with umpire as a submodule?

CMakeLists.txt
adayton1: This is actually a cmake thing. If you specify chai_DIR and call find_package(chai), it will look for chaiConfig.cmake or chai-config.cmake in the directory pointed to by chai_DIR. In this situation it is case sensitive.

cmake/SetupLibraries.cmake
adayton1: It should work to use an external library in place of all the submodules. The paths for umpire_DIR, chai_DIR, etc..., need to actually point to the directory containing umpire-config.cmake, chaiConfig.cmake, etc....

CMakeLists.txt
adayton1: `make install` in CHAI creates chaiConfig.cmake instead of CHAIConfig.cmake, so we need to use the lower case version.

cmake/SetupLibraries.cmake
dtaller: Do you think it would make sense to have CHAI_DIR etc for external users to be consistent with how other options and other packages work, and then copy those values to chai_DIR internally?

adayton1:
I merged in develop.

CMakeLists.txt
dtaller: Do you think it would make sense to have CHAI_DIR etc for external users to be consistent with how other options and other packages work, and then copy those values to chai_DIR internally?

cmake/SetupLibraries.cmake
adayton1: I've made it so we can use CHAI_DIR or chai_DIR.

CMakeLists.txt
adayton1: I've made it so we can use CHAI_DIR or chai_DIR.

src/care/host_device_ptr.h
```diff
@@ -398,6 +398,7 @@ namespace care {
       }
    }; // class host_device_ptr
 
+#if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
    /////////////////////////////////////////////////////////////////////////////////
```
dtaller, line 402: Two things: 1) I wouldn't surround this by CARE_ENABLE_IMPLICIT_CONVERSIONS, this function could be used to print a host_device_ptr in any case
2) In the description below, could you please change "this line is needed for..." to "this function is needed for..."

src/care/host_ptr.h
```diff
@@ -69,12 +69,15 @@ namespace care {
                    typename std::enable_if<B, int>::type = 1>
          host_ptr<T>(host_ptr<T_non_const> const &ptr) noexcept : m_ptr(ptr.data()) {}
 
+         // TODO: We can probably just use .data in the future, but some
+         // packages that depend on CARE use an older version of CHAI.
+#if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
          ///
```
dtaller, line 75: I wonder if we really need both cases here? We could probably just use host_ptr(chai::ManagedArray<T> const &ptr) : m_ptr(ptr.data()) {} in both cases regardless of if implicit conversions are on and off, would work the same and be less maintenance. Just have a comment explaining why you have .data() instead

dtaller: Nevermind, missed the comment above

src/care/host_device_ptr.h
dtaller: Actually you can ignore (1), but add a TODO comment to remove the CARE_ENABLE_IMPLICIT_CONVERSIONS when chai is updated

src/care/host_ptr.h
adayton1: So the .data method is in CHAI develop, but there is no CHAI release that contains it. That's why I'm hesitant to unconditionally use it. Most people stick to a tagged release of CHAI, in which case using the .data method will cause build failures. I do have a comment on lines 72-74 about that. But using CARE_ENABLE_IMPLICIT_CONVERSIONS is probably not the right thing to do. I may have to do something based on the version number or use the getPointer method instead since that was present in v2.1.1 (the latest release).

adayton1: I decided to use .getPointer and added a comment about updating to .data when the next release of CHAI occurs.

src/care/host_device_ptr.h
adayton1: (1) I removed the macro guard but used .getPointer instead of .data since there's no tagged release of chai with the .data method in it yet.
(2) Fixed.

PR #20, opened 2020-07-14
feature/dtaller/disableimplicit

dtaller:
Adds option to disable implicit conversions between host_device_ptr and raw arrays. These often cause errors when data is copied to/from the device.

EDIT: After discussions with Alan, we decided to leave implicit conversions on for host_ptrs for the time being. That could be fixed in a future update. Allowing implicit conversions for host_ptr is less dangerous because we don't copy data from host/device.

This pull request is related to the pull request https://github.com/LLNL/CHAI/pull/125/files/6bd5fe7e506432defe70582d4bf53a87122f1199..7820b09e38cb28f58bd5fff839275004770f3ece .

### DIFF
--- a/src/care/host_device_ptr.h
+++ b/src/care/host_device_ptr.h
@@ -72,6 +72,7 @@ namespace care {
       ///
       CARE_HOST_DEVICE host_device_ptr<T>(std::nullptr_t from) noexcept : MA (from) {}
 
+#if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
       ///
       /// @author Peter Robinson
       ///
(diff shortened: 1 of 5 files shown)


src/care/host_device_ptr.h
```diff
@@ -396,6 +398,16 @@ namespace care {
       }
    }; // class host_device_ptr
 
+
+   // Prints host_device data to the given ostream. 
+   // When ENABLE_IMPLICIT_CONVERSIONS is off, this line is needed for the line 'using stream_insertion_t = decltype(std::cout << std::declval<T>());' in
```
adayton1, line 403: I'm impressed you figured this out in the midst of all those template shenanigans.

adayton1: Just use doxygen style comments - I have it set up so that we actually can generate documentation from the code.

dtaller: Added the comments in doxygen style, no problem. 

It took me a long while staring blankly at strange template error messages to figure that one out :D . But yes, I learned a bit, and Google helped a bit :D . Glad that is resolved!

PR #22, opened 2020-07-21
First pass at continuous integration

adayton1:
Adds a new workflow script to trigger build and test on pull requests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 name: C/C++ CI
 
 on:
-  push:
-    branches: [ develop ]
   pull_request:
     branches: [ develop ]
 
@@ -20,6 +18,6 @@ jobs:
     - name: cmake
       run: cmake ../
     - name: make
-      run: make
+      run: make -j
     - name: make test
       run: make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,17 +10,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-    - name: init submodules
+    - name: Check out repository
+      uses: actions/checkout@v2
+    - name: Initialize submodules
       run: git submodule update --init
-    - name: build directory
+    - name: Make build directory
       run: mkdir build
-    - name: cmake
+    - name: Run cmake
       run: cmake ../
       working-directory: build
-    - name: make
+    - name: Run make
       run: make -j
       working-directory: build
-    - name: make test
+    - name: Run tests
       run: make test
       working-directory: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,13 @@ jobs:
     - name: init submodules
       run: git submodule update --init
     - name: build directory
-      run: mkdir build && cd build
+      run: mkdir build
     - name: cmake
       run: cmake ../
+      working_directory: build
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      run: mkdir build
+    - name: Run cmake
+      run: cmake ../
+      working-directory: build
+    - name: Run make
+      run: make -j
+      working-directory: build
+    - name: Run tests
+      run: make test
+      working-directory: build
```
dtaller, line 27: I would add another step for deletion of artifacts and files that this process creates. You could probably do a simple rm -rf for now, or if you want to be fancy there are options for dealing with artifacts like artifacts:expire_in : time in seconds before an artifact’s deletion ( https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html ).

src/care/host_device_ptr.h
```diff
@@ -409,10 +409,10 @@ namespace care {
    /// @return the output stream after printing
    /////////////////////////////////////////////////////////////////////////////////
    template<typename T>
-   std::ostream& operator<<(std::ostream& os, const host_device_ptr<T>& ptr) {
+   std::ostream& operator<<(std::ostream& os, host_device_ptr<T> ptr) {
```
dtaller, line 412: Add a TODO comment here saying that this can be reverted when we have a const version of these functions in chai

src/care/host_ptr.h
```diff
@@ -77,7 +77,7 @@ namespace care {
          ///
          /// Construct from chai::ManagedArray
          ///
-         host_ptr(chai::ManagedArray<T> const &ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
+         host_ptr(chai::ManagedArray<T> ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
```
dtaller, line 81: Likewise, add a TODO comment saying that this should be reverted to const in a future update of chai

src/care/host_ptr.h
```diff
@@ -86,7 +86,7 @@ namespace care {
          ///
          template <bool B = std::is_const<T>::value,
                    typename std::enable_if<B, int>::type = 1>
-         host_ptr<T>(chai::ManagedArray<T_non_const> const &ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
+         host_ptr<T>(chai::ManagedArray<T_non_const> ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
```
dtaller, line 90: Likewise, add a TODO comment saying that this should be reverted to const in a future update of chai . That way, we can just grep TODO when we next update chai and not miss anything

.github/workflows/ci.yml
adayton1: So github's process may be different than gitlab. But in either case I think everything in the working directory is deleted by default (including the cloned repository and it's subdirectories, which would include the created build directory). I think you specify artifacts if you want things to stay around. But just to be safe I added a remove of the build directory.

src/care/host_device_ptr.h
adayton1: Done.

src/care/host_ptr.h
adayton1: Done.

src/care/host_ptr.h
adayton1: Done.

PR #23, opened 2020-07-22
temporary fix for implicit cast build

dtaller:
In local pointer, we relied on implicit cast in constructor

### DIFF
--- a/src/care/local_ptr.h
+++ b/src/care/local_ptr.h
@@ -85,11 +85,15 @@ namespace care {
          CARE_HOST_DEVICE local_ptr<T>(host_ptr<T_non_const> const &ptr) noexcept : m_ptr(ptr) {}
 
          ///
+         /// TODO: When CHAI is updated, change the following constructors to ptr.data() instead of ptr.getPointer()
+         /// 
+
+         ///
          /// @author Peter Robinson
          ///
          /// Construct from chai::ManagedArray
          ///
-         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T> const &ptr) : m_ptr(ptr) {}
+         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T> const &ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
 
          ///
          /// @author Peter Robinson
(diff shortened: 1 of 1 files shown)


src/care/local_ptr.h
```diff
@@ -84,12 +84,16 @@ namespace care {
...
+         /// TODO: When CHAI is updated, change the following constructors to ptr.data() instead of ptr.getPointer()
+         /// 
+
          ///
          /// @author Peter Robinson
          ///
          /// Construct from chai::ManagedArray
          ///
-         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T> const &ptr) : m_ptr(ptr) {}
+         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T> const &ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
```
adayton1, line 96: Because this is a HOST_DEVICE function, you should use .data() with no parameters.

src/care/local_ptr.h
```diff
@@ -98,7 +102,7 @@ namespace care {
          ///
          template <bool B = std::is_const<T>::value,
                    typename std::enable_if<B, int>::type = 1>
-         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T_non_const> const &ptr) : m_ptr(ptr) {}
+         CARE_HOST_DEVICE local_ptr<T>(chai::ManagedArray<T_non_const> const &ptr) : m_ptr(ptr.getPointer(chai::CPU)) {}
```
adayton1, line 105: Same here.

src/care/local_ptr.h
dtaller: Yes, that is true, but it will probably have to wait until CHAI is updated for that to happen. I will hold off on this for the time being

src/care/local_ptr.h
```diff
@@ -84,12 +84,16 @@ namespace care {
                    typename std::enable_if<B, int>::type = 1>
          CARE_HOST_DEVICE local_ptr<T>(host_ptr<T_non_const> const &ptr) noexcept : m_ptr(ptr) {}
 
+         ///
```
adayton1, line 87: Remove this comment and I will merge in your pull request.

PR #25, opened 2020-07-23
some fixes for disable implicit conversions

dtaller:


src/care/KeyValueSorter.h
```diff
@@ -1135,15 +1135,15 @@ class KeyValueSorter<T, RAJA::seq_exec> {
             std::sort(rawData, rawData + lsize, cmpKeys<T>);
 
             // Reallocate memory
-            if (m_keyValues) {
+            if (m_keyValues != nullptr) {
```
adayton1, line 1138: The `if (ptr)` pattern should not need to be changed.

src/care/KeyValueSorter.h
```diff
@@ -1206,7 +1206,7 @@ class KeyValueSorter<T, RAJA::seq_exec> {
       /// @return whether keys are allocated
       ///////////////////////////////////////////////////////////////////////////
       bool keysAllocated() const {
-         return m_keys ;
+         return m_keys != nullptr;
```
adayton1, line 1209: This is the pattern that is fine to change. I'm fine with either what you have here or `return (bool) m_keys`. What you have here is probably actually better (easier to see what is going on).

dtaller: I kept it as != nullptr, sounds good

src/care/KeyValueSorter.h
dtaller: Reverted

PR #26, opened 2020-07-24
yet more fixes for disbaling implicit casts

dtaller:
I made a version of binary search and check sorted that can take in a host device pointer (without an implicit cast). Some other minor fixes

### DIFF
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -183,7 +183,7 @@ CARE_HOST_DEVICE bool checkSorted(const T* array, const int len,
 
 #if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
 template <typename T>
-CARE_HOST_DEVICE bool checkSorted(const care::host_device_ptr<T>& array, const int len,
+CARE_HOST_DEVICE bool checkSorted(const care::host_device_ptr<const T>& array, const int len,
                                   const char* name, const char* argname,
                                   const bool allowDuplicates = false);
 #endif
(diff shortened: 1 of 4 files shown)


src/care/array_utils.h
```diff
@@ -181,6 +181,13 @@ CARE_HOST_DEVICE bool checkSorted(const T* array, const int len,
                                   const bool allowDuplicates = false);
 
 
+#if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
```
adayton1, line 184: So if we update CHAI, this macro guard will be unnecessary, right?

src/care/array_utils.h
```diff
@@ -610,6 +633,26 @@ CARE_HOST_DEVICE inline int BinarySearch(const T *map, const int start,
...
+template<typename mapType>
+CARE_HOST_DEVICE inline int BinarySearch(const care::host_device_ptr<const mapType>& map, const int start,
+                                         const int mapSize, const mapType num,
+                                         bool returnUpperBound)
+{
+#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__) 
+   return BinarySearch<const mapType>(map.data(), start, mapSize, num, returnUpperBound);
+#else
+   int result = -1;
+   LOOP_SEQUENTIAL_REF(i, 0, 1, result) {
```
adayton1, line 646: I don't see why you need separate behavior for host vs. device. This function should just be a wrapper, so `return BinarySearch<const mapType>(map.data(), start, mapSize, num, returnUpperBound);` should be sufficient.

dtaller: Fixed

src/care/array_utils.h
dtaller: Yes, I added a comment to revisit this when CHAI is updated. It should not be needed if everything is consistent

adayton1: I'm thinking we should just update chai so we don't have to worry about this.

dtaller: Ok. I just double checked and got rid of the guard. care compiles with both implicit conversions on and off. I don't think it will compile with the ale3d develop version of chai (which is what I was using to build ale3d), so both of those will have to updated at the same time for things to work with ale3d.

dtaller: Wait just a sec, don't approve yet

dtaller: Ok, ready for another look after tests pass. Should be fine to get rid of guard now

adayton1: Looks good. Yeah, we should definitely update CHAI at the same time as CARE in the main src.

PR #28, opened 2020-08-08
add const constructors to KeyValueSorter

dtaller:


src/care/KeyValueSorter.h
```diff
@@ -223,6 +223,24 @@ class KeyValueSorter<T, RAJADeviceExec> {
...
+      /// @param[in] arr - The managed array to copy elements from
+      /// @return a KeyValueSorter instance
+      ///////////////////////////////////////////////////////////////////////////
+      KeyValueSorter<T, RAJADeviceExec>(const size_t len, const host_device_ptr<const T>& arr)
+      : m_len(len)
+      , m_ownsPointers(true)
+      , m_keys(len, "m_keys")
+      , m_values(len, "m_values")
+      {
+         setFromArray(len, arr.data());
```
adayton1, line 241: There is an overload of setFromArray that takes a `const host_device_ptr<T>&` as an argument. But the method never writes to it, so you should be able to make the argument a `const host_device_ptr<const T>&` and not have to use `.data()` here since that results in an extra copy to the host.

src/care/KeyValueSorter.h
```diff
@@ -223,6 +223,24 @@ class KeyValueSorter<T, RAJADeviceExec> {
...
+      ///////////////////////////////////////////////////////////////////////////
+      /// @author Alan Dayton
+      /// @brief Constructor
+      /// Allocates space and initializes the KeyValueSorter by copying
+      ///    elements and ordering from the given managed array
+      /// @param[in] len - The number of elements to allocate space for
+      /// @param[in] arr - The managed array to copy elements from
+      /// @return a KeyValueSorter instance
+      ///////////////////////////////////////////////////////////////////////////
+      KeyValueSorter<T, RAJADeviceExec>(const size_t len, const host_device_ptr<const T>& arr)
```
adayton1, line 235: You should simply be able to update the above constructor to accept `const host_device_ptr<const T>&` rather than adding another constructor.

src/care/KeyValueSorter.h
```diff
@@ -757,6 +775,25 @@ class KeyValueSorter<T, RAJA::seq_exec> {
...
+      ///////////////////////////////////////////////////////////////////////////
+      /// @author Alan Dayton
+      /// @brief Constructor
+      /// Allocates space and initializes the KeyValueSorter by copying
+      ///    elements and ordering from the given managed array
+      /// @param[in] len - The number of elements to allocate space for
+      /// @param[in] arr - The managed array to copy elements from
+      /// @return a KeyValueSorter instance
+      ///////////////////////////////////////////////////////////////////////////
+      KeyValueSorter<T, RAJA::seq_exec>(const size_t len, const host_device_ptr<const T>& arr)
```
adayton1, line 787: You should simply be able to update the above constructor to accept `const host_device_ptr<const T>&` rather than adding another constructor.

src/care/KeyValueSorter.h
adayton1: If the compiler doesn't like `const T` when `T` is already const (i.e. if `T` is `const real8`, I don't know how it would handle `const const real8`), then you could use type traits: `using T_const = typename std::add_const<T>::type;` Then you should be able to use `const host_device_ptr<T_const>&`. You'd have to include the <type_traits> header in this case.

dtaller: const host_device_ptr<const T>& worked fine, changed setFromArray to be consistent

src/care/KeyValueSorter.h
dtaller: Done, changed setFromArray to have the same signature also

src/care/KeyValueSorter.h
dtaller: Done

PR #34, opened 2020-08-21
Bugfixes and regression tests

adayton1:
Fixes several issues due to implicit casting and adds regression tests to make sure those issues don't come back.


reproducers/ReproducerManagedPtr.cpp
```diff
@@ -0,0 +1,76 @@
...
+   private:
+      int* m_data;
+      care::host_device_ptr<int> m_managedData;
+};
+
+
+int main(int, char**) {
+   // Set up data
+   int length = 10;
+   int data[10] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
dtaller, line 57: This is not a big deal, but you can just do int data[10] = {0} and then all ten elements would be 0. Would also work

src/care/host_device_ptr.h
```diff
@@ -422,6 +438,17 @@ namespace care {
    void using_host_device_ptr_outside_of_raja_loop_not_allowed(T foo); 
 } // namespace care
 
+#if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
+
+// TODO: Declaring these functions causes problems with a project that depends on CARE
```
dtaller, line 443: Thanks for figuring this out

test/TestKeyValueSorter.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+// other library headers
+#include "gtest/gtest.h"
+
+// care headers
+#include "care/array.h"
+#include "care/care.h"
+#include "care/KeyValueSorter.h"
+
+TEST(KeyValueSorter, SizeConstructor)
```
dtaller, line 20: Thanks for adding more tests

adayton1: This definitely made it easier to test what fixes were needed when implicit casts were enabled.

src/care/host_device_ptr.h
adayton1: No problem. I've got some ideas for a better solution, but they should be in a separate pull request.

reproducers/ReproducerManagedPtr.cpp
adayton1: Actually, that pattern usually leads to people misunderstanding what is happening. If I did `int data[10] = {1};` most people would think that it would be all filled with 1s, but actually there would be only one 1 and nine 0s. I believe I could do `int data[10] = {};` but that isn't necessarily clear, either.

PR #35, opened 2020-08-24
adds a fusible phase macro 

robinson96:
to allow fusion of multiple steps that need to be executed in a particular order.

Defines the following macros:

FUSIBLE_LOOP_PHASE(INDEX,START,END,PRIORITY)
FUSIBLE_LOOP_PHASE_END
a new loop type, labeled by a priority (double), that indicates order of execution that phases should be flushed in. 

FUSIBLE_PHASE_RESET
tells FusibleActionsObserver when a new iteration over phases occurs.


src/care/LoopFuser.cpp
```diff
@@ -34,24 +34,23 @@ LoopFuser::LoopFuser() :
...
    static LoopFuser * instance = nullptr;
    if (instance == nullptr) {
       instance = new LoopFuser();
-
-      // Supports fusing up to 1M loops of average lambda size of 256 bytes
-      // will flush if we exceed the 1M count or if the lambda size requirements
-      // are exceeded.
-      instance->reserve(1024*1024);
-      instance->reserve_lambda_buffer(256*1024*1024);
+      defaultObserver->registerFusedActions(instance, -FLT_MAX/2);
```
liu15, line 53: why is this value different from the default value?

robinson96: this is to avoid the default actions triggering the warning of registering a priority after / the same as the previous priority.

src/care/LoopFuser.h
```diff
@@ -339,6 +339,266 @@ class SerializableDeviceLambda {
...
+   ///////////////////////////////////////////////////////////////////////////
+   template <typename T>
+   void registerFree(care::host_device_ptr<T> & array) {
+      m_to_be_freed.push_back(reinterpret_cast<care::host_device_ptr<char> &>(array));
+   }
+
+   protected:
+      std::unordered_map<FusedActions *, double> m_fused_actions;
+      std::map<double, FusedActions *> m_fused_action_order;
+      double m_last_insert_priority = -1.0;
```
liu15, line 593: The default value is set here and in the only constructor.  Should we get rid of one of them?

src/care/LoopFuser.cpp
liu15: A comment here would be useful.

Is the intent here to always be greater than the existing one?  If we call getInstance(), delete it, and call getInstance() again we will get the warning, right?

src/care/LoopFuser.h
```diff
@@ -339,6 +339,266 @@ class SerializableDeviceLambda {
...
+            actions->flushActions(async);
+         }
+      }
+      for (auto & array: m_to_be_freed) {
+         array.free();
+      }
+      m_to_be_freed.clear();
+   }
+
+   void registerFusedActions(FusedActions * actions, double priority) {
```
adayton1, line 539: Do we actually want all these methods to be inline?

src/care/LoopFuser.cpp
robinson96: yes.

src/care/LoopFuser.h
robinson96: done

src/care/LoopFuser.h
robinson96: done

test/LoopFuserTest.cxx
```diff
@@ -581,9 +582,107 @@ GPU_TEST(fusible_scan_custom, basic_fusible_scan_custom) {
...
+   care::host_device_ptr<int> B3(arrSize);
+   care::host_device_ptr<int> C1(arrSize);
+   care::host_device_ptr<int> C2(arrSize);
+   care::host_device_ptr<int> C3(arrSize);
+
+   care::host_device_ptr<int> As[] = {A1,A2,A3};
+   care::host_device_ptr<int> Bs[] = {B1,B2,B3};
+   care::host_device_ptr<int> Cs[] = {C1,C2,C3};
+
+   printf(" starting fusible_phases\n");
```
adayton1, line 602: Remove print statement from test.

src/care/LoopFuser.h
```diff
@@ -733,37 +983,62 @@ void LoopFuser::registerFree(care::host_device_ptr<T> & array) {
...
 // in opt, non cuda builds, never start recording
-#define FUSIBLE_LOOPS_START { \
-   auto __fuser__ = LoopFuser::getInstance(); \
-   __fuser__->setScan(false); \
-   __fuser__->setCountsToOffsetsScan(false); \
+#define FUSIBLE_LOOPS_START \
+{ \
+   static FusedActionsObserver * __phase_observer = new FusedActionsObserver(); \
+   for ( FusedActions * __fuser__ : {static_cast<FusedActions *>(LoopFuser::getInstance()), static_cast<FusedActions *>(__phase_observer)}) { \
+      __fuser__->stopRecording(); \
```
adayton1, line 1031: Should this be startRecording?

src/care/LoopFuser.h
```diff
@@ -814,10 +1118,41 @@ void LoopFuser::registerFree(care::host_device_ptr<T> & array) {
       0, 1, __fusible_scan_pos__, \
       [=] FUSIBLE_DEVICE(int, bool, int, int, int)->bool { return true; }, \
       [=] FUSIBLE_DEVICE(int, bool, int, int, int)->int {
+
+#define FUSIBLE_LOOP_PHASE(INDEX, START, END, PRIORITY) { \
+   if (END > START) { \
```
adayton1, line 1123: You use very similar code multiple times. Is there a way to reuse the same code without obscuring everything? I think that might make it less error prone in the future if we adjust these macros.

src/care/LoopFuser.h
adayton1: I was really asking if having them inline is what we actually want. I wonder if having more loop fuser stuff implemented in the source file might reduce the compile times in places that include the loop fuser header.

robinson96: that's a good question. Can we make a ticket and pursue this as a follow up branch?

src/care/LoopFuser.h
robinson96: no, with non-gpu, non CARE_DEBUG builds we default to executing loops as we encounter them (so we force recording to be off).

src/care/LoopFuser.h
robinson96: just pushed some changes that take a pass at this

src/care/LoopFuser.h
```diff
@@ -690,8 +920,28 @@ void LoopFuser::registerAction(int start, int end, int &start_pos, Conditional &
...
          ++m_action_count;
+
+         if (m_action_count == m_reserved) {
+#ifdef FUSER_VERBOSE
+            printf("hit reserved flushActions\n");
+#endif
+            flushActions();
+         }
+         // flush if we are approaching our buffer allocation - we add some fuzz here because we do not know
+         // a priori what the next lamba size is going to be, but we need to ensure a flush so the
```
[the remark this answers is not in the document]
robinson96: fixed

src/care/LoopFuser.h
```diff
@@ -339,15 +340,275 @@ class SerializableDeviceLambda {
...
+public:
+   FusedActions() = default;
+   ///////////////////////////////////////////////////////////////////////////
+   /// @brief starts recording. If recording is stopped, registerAction calls will
+   ///        execute the lambda immediately. If recording has started, they will
+   ///        be gathered up until flushed either by filling up our buffer or via
+   ///        a flush call.
+   ///////////////////////////////////////////////////////////////////////////
+   virtual void startRecording() {
+#ifndef FUSIBLE_LOOPS_DISABLE
```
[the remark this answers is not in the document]
robinson96: this is probably rarely used, but the idea behind this macro is to allow compilation unit control over whether the fusible loop infrastructure batches things or executes them as it sees them. CARE_ENABLE_LOOP_FUSER disables the entire infrastructure, converting the loops to vanilla STREAM / SCAN loops. The former can be used to ensure your're getting into registerFusedActions like you expect, but the latter is just pretending it doesn't exist. I think I will change this to CARE_FUSIBLE_LOOPS_DISABLE

src/care/LoopFuser.h
adayton1: Sure.

src/care/LoopFuser.h
adayton1: That looks better! Thanks!

src/care/LoopFuser.h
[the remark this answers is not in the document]
adayton1: OK.

test/LoopFuserTest.cxx
```diff
@@ -581,9 +582,119 @@ GPU_TEST(fusible_scan_custom, basic_fusible_scan_custom) {
...
+   care::host_device_ptr<int> C1(arrSize);
+   care::host_device_ptr<int> C2(arrSize);
+   care::host_device_ptr<int> C3(arrSize);
+
+   care::host_device_ptr<int> As[] = {A1,A2,A3};
+   care::host_device_ptr<int> Bs[] = {B1,B2,B3};
+   care::host_device_ptr<int> Cs[] = {C1,C2,C3};
+
+   /* OK, so this is a bit convoluted, but the idea is to have a test where both execution order
+    * matters, awe shuffle which order we introduce kernels, and we have some memory as to whether
```
adayton1, line 603: awe?

liu15: also exection two lines latter

test/LoopFuserTest.cxx
```diff
@@ -581,9 +582,119 @@ GPU_TEST(fusible_scan_custom, basic_fusible_scan_custom) {
...
+         else {
+            B[i] = (B[i] + 1)>>t;
+         }
+      } CARE_STREAM_LOOP_END
+      // check that no phases have been executed yet
+      CARE_SEQUENTIAL_LOOP(i, 0, arrSize) {
+         EXPECT_EQ(A[i], -2);
+         EXPECT_EQ(C[i], -2);
+      } CARE_SEQUENTIAL_LOOP_END
+      /* the captures and CHAI checks have already occured, the FUSIBLE_LOOPS_STOP
```
liu15, line 671: occurred

test/LoopFuserTest.cxx
adayton1: You misspelled later, Ben :)

test/LoopFuserTest.cxx
robinson96: fixed

test/LoopFuserTest.cxx
robinson96: fixed

PR #38, opened 2020-08-25
Fix ambiguous functions

adayton1:
Add an overload so that checkSorted is not ambiguous when CARE is configured with CARE_ENABLE_IMPLICIT_CONVERSIONS=ON.

### DIFF
--- a/src/care/array_utils.h
+++ b/src/care/array_utils.h
@@ -186,6 +186,15 @@ CARE_HOST_DEVICE bool checkSorted(const care::host_device_ptr<const T>& array, c
                                   const char* name, const char* argname,
                                   const bool allowDuplicates = false);
 
+#if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
+
+template <typename T>
+CARE_HOST_DEVICE bool checkSorted(const care::host_device_ptr<T>& array, const int len,
+                                  const char* name, const char* argname,
+                                  const bool allowDuplicates = false);
+
+#endif // defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
+
 ///////////////////////////////////////////////////////////////////////////
 /// @author Ben Liu, Peter Robinson, Alan Dayton
 /// @brief Checks whether an array of type T is sorted and optionally unique.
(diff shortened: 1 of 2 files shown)


src/care/array_utils.h
```diff
@@ -186,6 +186,15 @@ CARE_HOST_DEVICE bool checkSorted(const care::host_device_ptr<const T>& array, c
                                   const char* name, const char* argname,
                                   const bool allowDuplicates = false);
 
+#if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
+
+template <typename T>
```
dtaller, line 191: Does this need to be #if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS) protected? 

Could we just have both 
bool checkSorted(const care::host_device_ptr<T>& array
and also  
bool checkSorted(const care::host_device_ptr<const T>& array 
, without any #if defined(IMPLICIT) ?

adayton1: This function is only necessary if CARE_ENABLE_IMPLICIT_CONVERSIONS=ON, so I think it best to avoid code bloat and also unnecessary layers of function calls since they just make debugging harder.

dtaller: If that is the case, what do you think of other functions with the same paradigm? For example, right below checkSorted we have a similar paradigm for BinarySearch. Should you have the same #if statements there, or not or save that for a different pull request?

template<typename mapType>
CARE_HOST_DEVICE int BinarySearch(const care::host_device_ptr<const mapType>& map, const int start,
                                  const int mapSize, const mapType num,
                                  bool returnUpperBound = false) ;
template<typename mapType>
CARE_HOST_DEVICE int BinarySearch(const care::host_device_ptr<mapType>& map, const int start,
                                  const int mapSize, const mapType num,
                                  bool returnUpperBound = false) ;

adayton1: I've added an issue for fixing those. I think that should go in a separate pull request.

dtaller: Ok, no problem.

PR #48, opened 2020-09-08
Re-add CARE_LEGACY_COMPATIBILITY_MODE, change to allow vista to compile, openmp build fix

dtaller:
Re-added compatibility mode to the build system, made a fix for openmp builds with compatibility mode, and resolved issue with implicit conversions between host_device_ptrs and raw pointers as described below:

Previously, when implicit conversions from managed pointer to raw pointer were disabled, operations like the following would fail:

arrays[j][i];

for  host_device_ptr<host_device_ptr < const T> const > arrays

This is because of a const issue with operator[], which is now resolved

Basically, the first operator[] would return a const host_device_ptr , but since operator[] was not const then the second access was not allowed.


liu15 (approved):

liu15 (commented):
Does this build in LEGACY mode?

dtaller:
> Does this build in LEGACY mode?

I just added LEGACY mode to the build options (It must have been removed by mistake while refactoring). It builds in that mode, just checked.

src/care/host_device_ptr.h
```diff
@@ -202,11 +202,9 @@ namespace care {
       /// Return the value at the given index
       ///
       template<typename Idx>
-      inline CARE_HOST_DEVICE T& operator[](const Idx i)
-#ifndef CARE_LEGACY_COMPATIBILITY_MODE
-      const
-#endif
-      { return MA::operator[](i); }
+      inline CARE_HOST_DEVICE T& operator[](const Idx i) const {
```
adayton1, line 205: I believe this change should no longer be necessary if CARE_LEGACY_COMPATIBILITY_MODE is correctly defined now.

adayton1 (changes requested):

src/care/host_device_ptr.h
dtaller: It seems that CARE_LEGACY_COMPATIBILITY_MODE is designed to change other behavior, including loop fusion, "permuted connectivity", how a bunch of the care loops operate )see care.h and DefaulyMacros.h). Are you really ok with changing all of that behavior on all builds? 

I need this change to get the implicit conversion stuff working, which would apply to all builds. I was hesitant to change a bunch of other behaviors

adayton1: CARE_LEGACY_COMPATIBILITY_MODE is already (or we thought it was already) turned on for the normal icc toss release build. It should be turned off for all other builds. When it is on, the loop macros degrade to raw for loops, meaning there is no lambda capture, and the compiler can do a lot more optimization making things much faster. Also, we don't have to violate the normal const rules, which state that if you are returning a non-const reference the method should be non-const. Hence why this method is only const if we are not in legacy mode.

adayton1: I would undo those changes and try adding an overload:

#if defined(CARE_LEGACY_COMPATIBILITY_MODE)
inline CARE_HOST_DEVICE const T& operator[](const Idx i) const {
          return MA::operator[](i); 
}
#endif

src/care/DefaultMacros.h
```diff
@@ -60,7 +60,7 @@
 /// @arg[in] CHECK The variable to check that the start and end macros match
 ///
 ////////////////////////////////////////////////////////////////////////////////
-#define CARE_CHECKED_FOR_LOOP_START(INDEX, START_INDEX, END_INDEX, CHECK) for (int INDEX = START_INDEX; INDEX < END_INDEX; ++INDEX) CARE_NEST_BEGIN(CHECK)
+#define CARE_CHECKED_FOR_LOOP_START(INDEX, START_INDEX, END_INDEX, CHECK) for (int INDEX = START_INDEX; INDEX < (int)END_INDEX; ++INDEX) CARE_NEST_BEGIN(CHECK)
```
dtaller, line 63: In ale3d, the users have added some loops where END_INDEX is a size_t ( vector.size() ). That is what motivated this cast to int

src/care/host_device_ptr.h
dtaller: Undid change, added your bit of code. We will talk Monday

src/care/host_device_ptr.h
```diff
@@ -208,6 +208,13 @@ namespace care {
 #endif
       { return MA::operator[](i); }
 
+#if defined(CARE_LEGACY_COMPATIBILITY_MODE)
+     template<typename Idx>
+     inline CARE_HOST_DEVICE T& operator[](const Idx i) const {
```
adayton1, line 213: This should return const T&

src/care/DefaultMacros.h
adayton1: Without this cast, is there a compiler warning/error?

adayton1: And does this only show up in the legacy build?

dtaller: Yes, in care compatibility mode it errored out, only in legacy. I was thinking of perhaps changing this to something fancy rather than a cast (I don't know if something like decltype could determine if we need a size_t or an int?), but I have not tried that yet.

Right now I'm changing things in my ale3d pull request to get the LEGACY flag passed to care properly, so please don't approve or merge this in until I have that resolved.

src/care/host_device_ptr.h
dtaller: I will probably kill this anyway after Peter said that was ok in the meeting.

But in any case, when I had const I still got some build issues where people tried to do arr[i][j] = stuff;, and there were const issues (I think the first operator[] returned a const which made the assignement impossible).

dtaller: Right now I'm changing things in my ale3d pull request to get the LEGACY flag passed to care properly, so please don't approve or merge this in until I have that resolved. There might be some other minor changes here, not sure

dtaller: Hi. I will keep this function. Please take another look, and tell me whether you want me to put the const in (in spite of issues when compatibility mode is turned on in ale3d)

src/care/host_device_ptr.h
dtaller: You can take another look. I will keep LEGACY mode disabled in ale3d for now; that will be a separate pull request

src/care/host_device_ptr.h
adayton1: You can leave this as is.

src/care/DefaultMacros.h
adayton1: Can you undo this change for now?

adayton1: Nevermind, we assume int anyway.

adayton1 (approved):
LGTM

PR #51, opened 2020-09-10
Disable managed_ptr stuff in the HIP build

adayton1:
Virtual functions are not support in HIP device calls, so managed_ptr must be disabled.

### DIFF
--- a/src/care/host_device_ptr.h
+++ b/src/care/host_device_ptr.h
@@ -438,6 +438,7 @@ namespace care {
    void using_host_device_ptr_outside_of_raja_loop_not_allowed(T foo); 
 } // namespace care
 
+#if defined(CARE_ENABLE_MANAGED_PTR)
 #if !defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
 
 // TODO: Declaring these functions causes problems with a project that depends on CARE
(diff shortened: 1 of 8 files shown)


robinson96 (approved):
LGTM

src/care/host_device_ptr.h
```diff
@@ -438,6 +438,7 @@ namespace care {
    void using_host_device_ptr_outside_of_raja_loop_not_allowed(T foo); 
 } // namespace care
 
+#if defined(CARE_ENABLE_MANAGED_PTR)
```
dtaller, line 441: Sorry, I may be missing something, but why are you putting this here? We are declaring getRawPointers with

care::host_device_ptr<T>

not managed_ptr , so it should be a different case?

adayton1: The original getRawPointers functions are defined in managed_ptr.hpp. So if managed_ptr support is disabled in CARE, there is no reason for this function.

dtaller:
> Virtual functions are not support in HIP device calls, so managed_ptr must be disabled.


Do we know if that will be changed/fixed for future versions of HIP? We should ask Kostas

dtaller (approved):
Looks good, just ask Kostas if virtual functions will be ok on the device in the future

PR #57, opened 2020-09-15
Remove old RAJA artifacts

adayton1:
We have lots of old RAJA defines that should actually be build configuration options of RAJA or simply do not exist anymore. This is an attempt to clean those up.


adayton1:
At some point along the line RAJA switched to the RAJA_ENABLE_OPENMP macro rather than RAJA_USE_OPENMP.

liu15 (approved):

PR #58, opened 2020-09-15
Remove more old RAJA artifacts

adayton1:
Removes the RAJA_COMPILER_ defines since these are handled in RAJA/config.hpp.


liu15 (approved):

robinson96 (commented):
LGTM

PR #60, opened 2020-09-16
Cleans up care.h

adayton1:
This pull request cleans up care.h. It moves some things into new headers, other things into existing headers, and removes some macros.


adayton1:
Using CARE_GPUCC instead of RAJA_GPU_ACTIVE ensures that the headers will be the same across translation units. I believe in theory it is fine to have the headers be different as long as no definitions are conflicting (and in these cases I think we just add additional definitions so nothing should be conflicting). But who knows if theory actually works in this case? And it might mess up debuggers.

adayton1:
This pull request will also enable us to only include the headers we care about rather than getting the entirety of CARE.

liu15 (approved):

PR #64, opened 2020-09-18
update chai

dtaller:
Updated chai

Some tests in test/TestAlgorithm seem to fail (produce wrong results). Needs to be investigated. Was an issue prior to this chai library update. Needs to be investigated

### DIFF
--- a/src/care/KeyValueSorter.h
+++ b/src/care/KeyValueSorter.h
@@ -783,7 +783,7 @@ class KeyValueSorter<T, RAJA::seq_exec> {
       }
 
 #if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
-      
+
       ///////////////////////////////////////////////////////////////////////////
       /// @author Alan Dayton
       ///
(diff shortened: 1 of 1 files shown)


src/care/KeyValueSorter.h
```diff
@@ -223,8 +223,6 @@ class KeyValueSorter<T, RAJADeviceExec> {
          setFromArray(len, arr);
       }
 
-#if defined(CARE_ENABLE_IMPLICIT_CONVERSIONS)
```
adayton1, line 226: Why were these removed?

dtaller: Needed to compile both with and without implicit conversions disabled. I tried both, and I needed this  when CARE_ENABLE_IMPLICIT_CONVERSIONS was OFF also

adayton1: The actual issue is that implicit conversions need to be turned off in CHAI as well, and they haven't been. I've got a fix that will take care of this, so you can revert these changes.

dtaller: Ok, sounds good. I reverted the changes. Just make sure that things work with implicit conversions both on and off in you pull request afterwards.

adayton1: It works. I double checked.

adayton1 (approved):
LGTM

PR #73, opened 2020-10-06
Fixes inconsistent requirements in IntersectArrays

adayton1:
These fixes are courtesy of Ben Liu. The parameters of IntersectArrays are now correctly documented. What once was inconsistently considered the size of an array is now treated as the length of the array segment under consideration. The tests are updated to fit with these clarified specifications.


liu15 (approved):

dtaller (approved):

dtaller (approved):
With the build error fixed, should be fine now. Thanks Ben and Alan

PR #75, opened 2020-10-13
Rename ArrayFill to fill_n

adayton1:
Begin making the algorithms look more like the std library.


dtaller (approved):
Looks good to me

agcapps (approved):
Thanks, @adayton1 !

PR #79, opened 2020-10-20
windows and configuration build fixes for compatibility with raja 0.12

robinson96:
Addresses some build fixes to be compatible with RAJA 0.12.


src/care/CHAICallback.cpp
```diff
@@ -294,7 +296,9 @@ namespace care {
                   }
 
                   if (s_log_data > 0) {
+#ifndef CARE_DISABLE_RAJAPLUGIN
```
adayton1, line 299: Might as well have this outside the if statement. I think there are several places like this.

adayton1 (approved):

PR #80, opened 2020-10-26
fix CPU only Loop Fuser tests.

robinson96:
updates LoopFuserTest to use CARE_DEBUG instead of DEBUG to force enablement of LoopFuser on CPU builds.


adayton1 (approved):
LGTM

PR #81, opened 2020-10-28
Adds run time selectable policies for forall

adayton1:
- Adds an enum with a limited number of run time selectable policies
- Adds an overload of forall that takes this enum as an argument
- Uses the existing CARE_LOOP as a mechanism for calling the new overload of forall (also works with the compile time policies)
- CARE_LOOP uses CARE_HOST_DEVICE for the lambda, which means if CUDA or HIP is enabled these lambdas are always extended `__host__ __device__` lambdas (no dependency on GPU_ACTIVE). Otherwise they are always on the host only.

### DIFF
--- a/src/care/forall.h
+++ b/src/care/forall.h
@@ -347,6 +347,55 @@ namespace care {
       }
 #endif
    }
+
+   ////////////////////////////////////////////////////////////////////////////////
+   ///
+   /// @author Alan Dayton
+   ///
+   /// @brief Execute on the host. This specialization is needed for clang-query.
+   ///
+   /// @arg[in] sequential Used to choose this overload of forall
+   /// @arg[in] fileName The name of the file where this function is called
+   /// @arg[in] lineNumber The line number in the file where this function is called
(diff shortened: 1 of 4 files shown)


adayton1:
There are also safe fallbacks if you use gpu or openmp and do not have the appropriate build - in those cases it will just run sequentially on the host.

src/care/forall.h
```diff
@@ -347,6 +347,46 @@ namespace care {
       }
 #endif
    }
+
+   ////////////////////////////////////////////////////////////////////////////////
+   ///
+   /// @author Alan Dayton
+   ///
+   /// @brief Execute on the host. This specialization is needed for clang-query.
```
robinson96, line 355: these comments are out of date

robinson96 (changes requested):

src/care/forall.h
adayton1: Fixed.

robinson96 (approved):

PR #86, opened 2020-11-04
First pass at adding gitlab CI

adayton1:


adayton1:
If you haven't already logged on to the CZ gitlab instance, please do so: https://lc.llnl.gov/gitlab. Then I can add you to the mirror of the CARE repo there. Near the bottom of this page: https://lc.llnl.gov/confluence/display/GITLAB/GitLab+CI it sounds like you need to add a comment like "LGTM" to actually trigger the gitlab CI. But I'm not sure if that will happen on this pull request or if it needs to be in develop first.

.gitlab/ci/helper.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+######################################################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC and other CARE developers.
+# See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+######################################################################################
+
+mkdir build
```
dtaller, line 10: If possible, I would change this to mkdir build-cuda or mkdir build-blueos-nvcc10.1.243_clang8.0.1.cmake or something more descriptive.

.gitlab-ci.yml
```diff
@@ -0,0 +1,20 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+######################################################################################
+
+variables:
+   GIT_SUBMODULE_STRATEGY: normal
+
+stages:
+   - build
+
+gpu:
```
dtaller, line 14: Would prefer this to be cuda or something more specific then gpu, because someday we might also have amd gpus

dtaller (commented):
Overall fine, tell me if you think my comments on name changes are too nitpicky or if they are reasonable

.gitlab/ci/helper.sh
adayton1: I changed it to build-cuda. When we actually have more than one build, I will definitely be more specific.

.gitlab-ci.yml
adayton1: Good point. I changed the name to cuda.

adayton1: I will leave more specific names for when we actually do more than one build.

dtaller (approved):
Thanks for the name changes. Looks good to me now!

.gitlab-ci.yml
dtaller: Thanks!

.gitlab/ci/helper.sh
dtaller: Thanks!

liu15 (approved):

PR #88, opened 2020-11-07
Remove STREAM_P loop macro

adayton1:
The STREAM_P macro is really not useful since we already have a SEQUENTIAL_P macro.


dtaller (approved):
Looks good to me

PR #89, opened 2020-11-12
Fixes for when implicit casts are disabled

adayton1:
* Adds cdata method to host_ptr and local_ptr
* Removes some implicit casts in CARE


liu15 (approved):

robinson96 (approved):
LGTM

PR #92, opened 2020-11-24
rename CUDAWatchpoint to GPUWatchpoint, fix header include

dtaller:
I need to include Cuda watch point for both the hip and cuda debug build, not just the cuda build as before

Also rename CudaWatchpoint to GPUWatchpoint

### DIFF
--- a/src/care/util.h
+++ b/src/care/util.h
@@ -15,6 +15,9 @@
 // Other library headers
 #if defined(__CUDACC__)
 #include "cuda.h"
+#endif
+
+#if defined(CARE_GPUCC)
 #if defined(CARE_DEBUG)
 #include "care/CUDAWatchpoint.h"
 #endif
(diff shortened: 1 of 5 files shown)


src/care/util.h
```diff
@@ -15,6 +15,9 @@
 // Other library headers
 #if defined(__CUDACC__)
 #include "cuda.h"
+#endif
+
+#if defined(CARE_GPUCC)
 #if defined(CARE_DEBUG)
 #include "care/CUDAWatchpoint.h"
```
adayton1, line 22: Should this file be renamed GPUWatchpoint.h? Does it have both HIP and CUDA code in it?

dtaller: That's a good point. Yes, it is designed for both. Will do

dtaller: Done.

adayton1 (approved):
LGTM

PR #100, opened 2020-12-15
Feature/benliu/care extern

liu15:


src/care/algorithm.h
```diff
@@ -1,258 +1,1685 @@
-//////////////////////////////////////////////////////////////////////////////////////
```
adayton1, line 1: This copyright notice needs to be added back in.

src/care/algorithm_decl.h
```diff
@@ -0,0 +1,334 @@
...
+
+// Other CARE headers
+#include "care/host_device_ptr.h"
+#include "care/host_ptr.h"
+#include "care/local_ptr.h"
+#include "care/policies.h"
+
+// Other library headers
+#ifdef __CUDACC__
+#include "cub/cub.cuh"
```
adayton1, line 22: I think this only needs to be included in algorithm.h

src/care/KeyValueSorter_decl.h
```diff
@@ -0,0 +1,1300 @@
...
+#include "care/config.h"
+
+// Other CARE headers
+#include "care/LoopFuser.h"
+#include "care/algorithm_decl.h"
+
+// Other library headers
+#ifdef CARE_GPUCC
+#ifdef __CUDACC__
+#include "cub/cub.cuh"
```
adayton1, line 21: Shouldn't some of these headers only be included in the implementation file?

adayton1 (changes requested):
I left a few comments.

adayton1 (approved):
LGTM

PR #102, opened 2020-12-16
minor fix for LoopFuser with hip enabled

dtaller:
cuda has the function cudaHostAlloc whereas hip prefers hipHostMalloc (note the extra M), and I am fixing that in one place.  

See also the review on spack to add this option, https://github.com/spack/spack/pull/20417


adayton1 (approved):
LGTM

PR #105, opened 2021-01-12
Scan extern

liu15:
- Refactor scan.h to use overloaded functions instead of inline templates
- Place scan functions inside care namespace
- Use CARE_GPUCC/RAJADeviceExec consistently instead of RAJA_PARALLEL_ACTIVE/RAJAExec for template instances


adayton1 (approved):
LGTM

PR #107, opened 2021-01-14
Bugfix/probinso/fusible scan debug

robinson96:
Use RAJA events to correctly prevent race condition associated with asynchronous flushes of LoopFuser

### DIFF
--- a/src/care/LoopFuser.h
+++ b/src/care/LoopFuser.h
@@ -774,8 +774,6 @@ class LoopFuser : public FusedActions {
        using EventType = RAJA::resources::Event;
        EventType m_wait_for_event;
 
-      //RAJA::resources::Event m_wait_for_event;
-
       ///
       /// whether we are waiting for an active event
       ///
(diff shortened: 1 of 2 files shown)


src/care/LoopFuser.h
```diff
@@ -776,6 +816,7 @@ void LoopFuser::registerAction(const char * fileName, int lineNumber, int start,
          if (verbose) {
             printf("%p: Registering action %i type %i with start %i and end %i\n", this, m_action_count, scan_type, start, end);
          }
+         waitIfNeeded();
```
MrBurmark, line 819: This may be more restrictive than necessary, you can wait until just before you assign the result of instantiate to m_cw or m_aw. That could let you enqueue new work in the pool while the old work is still running form the memory owned by the group.

robinson96: I would like to do this in a separate PR so I can more rigorously test the change in isolation.

adayton1 (approved):
LGTM

PR #109, opened 2021-01-20
Fix globalID instantiations (for GPU gid64 builds)

liu15:


robinson96 (approved):
LGTM

src/care/scan.cpp
```diff
@@ -39,21 +39,21 @@ void getFinalScanCount(chai::ManagedArray<int> scanvar, int length, int& scanCou
    scanCount = scanvar.pick(length);
 }
 
-#if CARE_HAVE_LLNL_GLOBALID
+#if CARE_HAVE_LLNL_GLOBALID && GLOBALID_IS_64BIT
```
adayton1, line 42: I guess I'm confused why we have this GLOBALID_IS_64BIT as part of the macro guard. Don't we want these instantiations regardless of whether GIDTYPE is 64 bits or a regular int?

liu15: GIDTYPE is the raw type, so it is either int or int64_t.  If it is int, then this version is a duplicate and the compiler complains.

adayton1: Ah, that makes more sense.

adayton1 (approved):
LGTM

PR #110, opened 2021-01-21
Feature/probinso/register binning 2

robinson96:
SUMMARY:
LoopFuser is now templated on max register count and a bin-specific function signature

Details:
all preexisting FUSIBLE_LOOP_* macros use default value of 256.
Can use new FUSIBLE_LOOP_*_R(REGISTER_COUNT) macros to select 32, 64,128, or 256 max register counts.
If different count values are used within the same FUSIBLE_LOOPS_START / FUSIBLE_LOOPS_STOP region, one kernel launch will occur for each value.

Attempts to use values other than 32, 64,128, or 256 will result in compiler errors.
Exceeding the max register count in a fused kernel will result in a linking error.
Using lower max count will result in better occupancy, improving performance for fused kernels with large numbers of threads by up to 8x.

### DIFF
--- a/src/care/LoopFuser.cpp
+++ b/src/care/LoopFuser.cpp
@@ -40,9 +40,10 @@ bool operator!=(umpire::TypedAllocator<T> const& lhs, umpire::TypedAllocator<U> 
 #include "care/scan.h"
 #include "care/Setup.h"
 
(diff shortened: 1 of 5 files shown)


adayton1 (approved):
LGTM

liu15 (approved):

PR #114, opened 2021-02-08
Fix order of registered frees in loop fuser to prevent frees occurrin…

liu15:
…g before all actions are flushed

### DIFF
--- a/src/care/LoopFuser.h
+++ b/src/care/LoopFuser.h
@@ -373,10 +373,14 @@ public:
    inline void flushActions(bool async, const char * fileName, int lineNumber) {
       for (auto & priority_action : m_fused_action_order) {
          FusedActions * const & actions = priority_action.second;
-         if (actions -> actionCount() > 0) {
+         if (actions->actionCount() > 0) {
             if (verbose) {
                printf("flushing actions at priority %g\n", priority_action.first);
             }
+            // We allow the Observer to flush actions asynchronously.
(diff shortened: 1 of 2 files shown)


robinson96 (approved):
LGTM

src/care/LoopFuser.h
```diff
@@ -374,9 +374,15 @@ class FusedActionsObserver : public FusedActions {
       for (auto & priority_action : m_fused_action_order) {
          FusedActions * const & actions = priority_action.second;
          if (actions -> actionCount() > 0) {
-            actions->flushActions(async, fileName, lineNumber);
+            if (verbose) {
+               printf("flushing actions at priority %g\n", priority_action.first);
+            }
+            actions->flushActions(true, fileName, lineNumber);
```
adayton1, line 380: Shouldn't you be using `async` instead of `true`?

liu15: @robinson96 was this change intentional or leftover from debugging?

adayton1: OK, that makes more sense.

adayton1 (approved):
LGTM

src/care/LoopFuser.h
adayton1: I would just add a comment to the code explaining this.

adayton1 (approved):
LGTM

PR #116, opened 2021-02-19
Rename EVERYWHERE_KERNEL to MANAGED_PTR_UPDATE; ensure gpu_sim mode o…

liu15:
…nly updates the managed_ptr once

### DIFF
--- a/src/care/forall.h
+++ b/src/care/forall.h
@@ -307,10 +307,15 @@ namespace care {
    ///
    /// @brief Loops over the given indices and calls the loop body with each index.
    ///        This overload is CHAI and RAJA aware. It sets the execution space
-   ///        accordingly and calls RAJA::forall. First executes on the device, and
-   ///        then on the host.
+   ///        accordingly and calls RAJA::forall. First executes on the host, and
+   ///        then on the device.
+   /// @note In GPU_SIM mode, this does not simulate the call on the device because
+   ///       there is only a single pointer for the managed_ptr in GPU_SIM mode.
+   ///       If managed_ptr is ever updated to use a separate space in GPU_SIM mode,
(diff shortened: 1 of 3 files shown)


dtaller (approved):
Looks good to me

src/care/forall.h
```diff
@@ -319,21 +324,19 @@ namespace care {
...
          chai::ArrayManager* threadRM = chai::ArrayManager::getInstance();
          threadRM->setExecutionSpace(chai::GPU);
 
-#if CARE_ENABLE_GPU_SIMULATION_MODE
-         RAJA::forall<RAJA::seq_exec>(RAJA::RangeSegment(start, end), body);
-#elif defined(__CUDACC__)
+#if defined(__CUDACC__)
          RAJA::forall< RAJA::cuda_exec<CARE_CUDA_BLOCK_SIZE, CARE_CUDA_ASYNC>>(RAJA::RangeSegment(start, end), body);
 #elif defined(__HIPCC__)
          RAJA::forall< RAJA::hip_exec<CARE_CUDA_BLOCK_SIZE, CARE_CUDA_ASYNC>>(RAJA::RangeSegment(start, end), body);
```
dtaller, line 342: This is not at all related to your PR at all, but looking back at this we probably should not be using a variable called CARE_CUDA_BLOCK_SIZE for both CUDA and HIP . I don't know if the right change is to re-name this for CARE_GPU_BLOCK_SIZE, or to make two different variables in case a different number os optimal for each type of GPU.

Also, we probably should be using RAJADeviceExec (which is equal to RAJA::cuda_exec or RAJA::hip_exec or RAJA::omp_parallel_for_exec depending on how we compile). I'll add issues about these things

adayton1 (approved):
LGTM

PR #119, opened 2021-03-08
Fix error messages for care::exclusive_scan

adayton1:
* If the size is 0, then a no-op should occur without a warning now
* Add a warning if inPlace is false, but outData is nullptr


dtaller (approved):
Looks good to me

liu15 (approved):

PR #123, opened 2021-03-18
Fix Loop Fuser Variant

dtaller:
Before this PR Loop fuser files and tests were built even if CARE_ENABLE_LOOP_FUSER is set to OFF. This fixes that build issue. 

With this PR, if CARE_ENABLE_LOOP_FUSER is set ON, then loop fuser functionality is exercised (whether set on CPU or GPU builds). If it is OFF, then loop fuser files and tests are not compiled

Attempting to exercise loop fuser functionality on the CPU builds results in some errors, so I ifdefed those tests out. That is an issue for a future PR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@
 cmake_minimum_required(VERSION 3.4)
 
 project(care LANGUAGES CXX)
+include(CMakeDependentOption)
 
 ################################
 # Build options
@@ -33,7 +34,7 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
-option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF)
+cmake_dependent_option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF "ENABLE_CUDA OR ENABLE_HIP" OFF)
 
 #  Known to not compile for Debug configurations, have low probability of working in 
 #  practice for Release / RelWithDebInfo configurations.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
+option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF)
+
 #  Known to not compile for Debug configurations, have low probability of working in 
 #  practice for Release / RelWithDebInfo configurations.
 option(CARE_ENABLE_FUSER_BIN_32 "Enable the 32 register fusible loop bin." OFF)
@@ -116,11 +118,8 @@ if (ENABLE_CUDA)
 
    # CUDA optional dependencies without submodules (not open source)
    set(NVTOOLSEXT_DIR "" CACHE PATH "Path to external NVTOOLSEXT")
-   # Optional features
-   set(CARE_ENABLE_LOOP_FUSER ON CACHE STRING "Enable the loop fuser")
-else()
-   set(CARE_ENABLE_LOOP_FUSER OFF CACHE STRING "Enable the loop fuser")
 endif()
+
 ################################
 # BLT
 ################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,6 @@
 cmake_minimum_required(VERSION 3.4)
 
 project(care LANGUAGES CXX)
-include(CMakeDependentOption)
 
 ################################
 # Build options
@@ -34,7 +33,7 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
-cmake_dependent_option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF "ENABLE_CUDA OR ENABLE_HIP" OFF)
+option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" ON)
 
 #  Known to not compile for Debug configurations, have low probability of working in 
 #  practice for Release / RelWithDebInfo configurations.
--- a/src/care/CMakeLists.txt
+++ b/src/care/CMakeLists.txt
@@ -38,7 +38,6 @@ set(care_headers
     KeyValueSorter_inst.h
     local_host_device_ptr.h
     local_ptr.h
-    LoopFuser.h
     managed_ptr.h
     openmp.h
     SortFuser.h
@@ -57,10 +56,14 @@ set(care_headers
 set(care_sources
     care.cpp
     CHAICallback.cpp
-    LoopFuser.cpp
     RAJAPlugin.cpp
     scan.cpp
     )
+
+if (CARE_ENABLE_LOOP_FUSER)
+   list(APPEND care_headers LoopFuser.h)
+   list(APPEND care_sources LoopFuser.cpp)
+endif()
 
 if(CARE_ENABLE_EXTERN_INSTANTIATE)
    list(APPEND care_headers care_inst.h)
--- a/test/Benchmarks.cpp
+++ b/test/Benchmarks.cpp
@@ -53,6 +53,8 @@ static unsigned int currentColor = 0;
 using namespace care;
 using int_ptr = host_device_ptr<int>;
 
+#if defined(CARE_GPUCC)
+//This test is not valid on the CPU since AllocationStrategy chai::PINNED does not exist for that build.
 CUDA_TEST(TestFuser, Initialization) {
    printf("Initializing\n");
    // initializing and allocate memory pools
@@ -68,7 +70,7 @@ CUDA_TEST(TestFuser, Initialization) {
    care::syncIfNeeded();
    printf("Initialized... Benchmarking Loop Fusion\n");
 }
-
+#endif //CARE_GPUCC
 
 CUDA_TEST(TestFuser, OneMillionSmallKernels) {
    int numLoops = 1000000;
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -153,15 +153,18 @@ target_include_directories(TestScan
 blt_add_test( NAME TestScan
               COMMAND TestScan )
 
-blt_add_executable( NAME Benchmarks
-                    SOURCES Benchmarks.cpp
-                    DEPENDS_ON ${care_test_dependencies} )
+if (CARE_ENABLE_LOOP_FUSER)
+   blt_add_executable( NAME Benchmarks
+                       SOURCES Benchmarks.cpp
+                       DEPENDS_ON ${care_test_dependencies} )
 
-target_include_directories(Benchmarks
-                           PRIVATE ${PROJECT_SOURCE_DIR}/src)
+   target_include_directories(Benchmarks
+                              PRIVATE ${PROJECT_SOURCE_DIR}/src)
 
-target_include_directories(Benchmarks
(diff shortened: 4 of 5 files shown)


test/CMakeLists.txt
```diff
@@ -32,31 +32,33 @@ target_include_directories(TestForall
 blt_add_test( NAME TestForall
               COMMAND TestForall )
 
-blt_add_executable( NAME TestLoopFuser
-                    SOURCES LoopFuserTest.cxx
-                    DEPENDS_ON ${care_test_dependencies} )
+if (CARE_ENABLE_LOOP_FUSER)
```
adayton1, line 35: At the very bottom of this file there is a test called Benchmarks. That also uses the loop fuser and needs to be protected. (Eventually we'll want to make it a proper benchmark, but that is outside the scope of this pull request.)

CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
+option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF)
```
adayton1, line 36: I think we should use cmake_dependent_option.

Add this at around line 11: `include(CMakeDependentOption)`

Then you can use something like this here: `cmake_dependent_option(CARE_ENABLE_LOOP_FUSER) "Enable the loop fuser" ON "ENABLE_CUDA" OFF)`

test/CMakeLists.txt
dtaller: Changed, thanks good catch

CMakeLists.txt
dtaller: Done.

test/CMakeLists.txt
```diff
@@ -151,15 +153,18 @@ target_include_directories(TestScan
 blt_add_test( NAME TestScan
               COMMAND TestScan )
 
-blt_add_executable( NAME Benchmarks
-                    SOURCES Benchmarks.cpp
-                    DEPENDS_ON ${care_test_dependencies} )
+if (CARE_ENABLE_LOOP_FUSER)
```
adayton1, line 156: Just move this test up by the other ones so at a glance it is obvious that it exercises the same functionality.

CMakeLists.txt
```diff
@@ -33,6 +34,8 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
+cmake_dependent_option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" OFF "ENABLE_CUDA OR ENABLE_HIP" OFF)
```
adayton1, line 37: Danny, please use the line I gave you. We want the loop fuser to default to on for the CUDA build, not the HIP build or any other build.

adayton1: Actually, maybe we shouldn't be using cmake_dependent_option. Looking at the documentation a bit more, I think this prevents the user from setting the option.

dtaller: Wouldn't your line though disable the option for HIP build? The previous email suggested that Peter wanted it available for HIP testing

dtaller: Ah, I see your new comment now, will revert.

test/CMakeLists.txt
dtaller: That makes sense, changed

adayton1 (approved):
LGTM

dtaller:
> LGTM

It looks like a test is failing, I must have messed things up when I moved stuff around somehow; just a sec

dtaller:
> > LGTM
> 
> It looks like a test is failing, I must have messed things up when I moved stuff around somehow; just a sec

After discussions with @adayton1 , we set LOOP_FUSER to be on by default for all builds (both host and device). 

Previous behavior: Loop fuser tests would get built on the host, but the relevant logic within LoopFuser.cpp was ifdefed out to avoid any errors

Current behavior: We can exercise the loop fuser behavior on the CPU and GPU, or we can disable the behavior in which case the loop fuser files and tests do not get compiled at all. Exercising the loop fuser behavior on the CPU results in errors because we sometimes reference GPU-only behaviors which are not legal

adayton1 (approved):
LGTM

test/CMakeLists.txt
adayton1: Thanks!

test/Benchmarks.cpp
```diff
@@ -53,6 +53,8 @@ static unsigned int currentColor = 0;
 using namespace care;
 using int_ptr = host_device_ptr<int>;
 
+#if defined(CARE_GPUCC)
+//This test is not valid on the CPU since AllocationStrategy chai::PINNED does not exist for that build.
```
liu15, line 57: Is this true even in GPU sim mode?

adayton1: No. Are you thinking this should be `#if defined(CARE_GPUCC) || defined(CARE_ENABLE_GPU_SIMULATION_MODE)`?

liu15 (approved):

test/Benchmarks.cpp
liu15: This and many other conditionals, I think.

dtaller: I tried recompiling as 

cmake .. -DCMAKE_CXX_COMPILER=/usr/tce/packages/intel/intel-19.0.4/bin/icpc -DCMAKE_BUILD_TYPE=Debug -DCARE_ENABLE_GPU_SIMULATION_MODE=ON

and I am still getting failures and segfaults in this test and other loop fuser tests, although sometimes in different places than before it seems. 

Previously, in LoopFuser.cpp everything was ifdefed out for non-GPU builds in favor of some dummy implementation, so I think this was not exercised even for GPU_SIMULATION

liu15: ALE3D builds with the following:
# Make sure the resource manager is on
set(CHAI_DISABLE_RM "OFF" CACHE STRING "")
# Enable pinned memory
set(CHAI_ENABLE_PINNED "ON" CACHE STRING "")
# Ensure Loop Fuser is turned on 
set(CARE_ENABLE_LOOP_FUSER "ON" CACHE STRING "")

# Make sure lambdas are turned on
set(ENABLE_LEGACY_CARE "0" CACHE STRING "")

# Make sure GPU simulation mode is turned on
set(CHAI_ENABLE_GPU_SIMULATION_MODE "ON" CACHE STRING "")
set(CARE_ENABLE_GPU_SIMULATION_MODE "ON" CACHE STRING "")


Does this mean that the loop fuser is still disabled for that build?

dtaller: Just a sec, I'll do a test on the recent CARE develop version (not my branch) with your options and test what happens

dtaller: In my ALE3D repo, I checked out care and its dependencies, and in the file care/src/care/LoopFuser.cpp I made the following changes to see if it gets built:

// CARE config header
#include "care/config.h"

#if CARE_ENABLE_LOOP_FUSER

//see if this triggers an error on GPU SIM mode...
oogabooga!!!

// Other CARE headers
#include "care/DefaultMacros.h"


And then I did the following:

make pmake-all VARIANT=gpu_sim

It seems like it was being triggered. So it seems like the ale3d version was fine. I'll investigate a bit

dtaller: I'll take a closer look after lunch, sorry for the confusion.

dtaller: Good news! I compiled this with the development CARE version (NOT my branch) using the command  cmake .. -DDCMAKE_C_COMPILER=/usr/tce/packages/intel/intel-19.0.4/bin/icc -DCMAKE_CXX_COMPILER=/usr/tce/packages/intel/intel-19.0.4/bin/icpc -DCHAI_DISABLE_RM=OFF -DCHAI_ENABLE_PINNED=ON -DCARE_ENABLE_LOOP_FUSER=ON -DENABLE_LEGACY_CARE=0 -DCHAI_ENABLE_GPU_SIMULATION_MODE=ON -DCARE_ENABLE_GPU_SIMULATION_MODE=ON . Hopefully that is the right command?

I got the exact same errors as with my branch and with the development CARE branch. Happily, this means that any issues with the GPU SIM mode were not introduced as a result of this PR, but were per-exisiting.

dtaller: I did some more investigating. I built care as part of ale3d (so using the version of care that is currently used in ale3d), building in gpu-sim mode and turning tests on (they were off by default).

It seems that the Benchmarks test is still failing in the version of CARE used by ale3d in GPU SIM mode - more evidence that this was a pre-existing issue not caused by this PR.

dtaller:
Hi. Don't merge this for now. because we have #include "care/LoopFuser.h" in various locations in vista and src dependencies, this change would break our current ale3d build. Maybe it makes sense to just require LoopFuser for all builds? We can discuss next week

liu15:
I think building with the current ALE3D configurations should be a prerequisite for this to go in.  The LoopFuser file should still be compiled but not do any fusing when the loop fuser is disabled.

The LoopFuser files are still needed so that CARE can provide a common API whether you use loop fuser or not.

test/Benchmarks.cpp
liu15: Ok, so the tests for GPU_SIM mode will need to be revisited elsewhere.

liu15 (approved):

CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ option(CARE_ENABLE_MANAGED_PTR "Enable managed_ptr aliases, tests, and reproduce
 option(CARE_DISABLE_RAJAPLUGIN "Disable use of the RAJA plugin. WILL ALSO DISABLE MEMORY MOTION." OFF)
 option(CARE_ENABLE_EXTERN_INSTANTIATE "Enable extern instantiation of template functions" OFF)
 option(CARE_ENABLE_GPU_SIMULATION_MODE "Enable GPU simulation mode." OFF )
+option(CARE_ENABLE_LOOP_FUSER "Enable the loop fuser" ON)
```
robinson96, line 36: I'm not 100% sure the default should be ON. I think the default makes sense to be off for CPU builds and ON for both HIP and CUDA builds.

src/care/CMakeLists.txt
```diff
@@ -38,7 +38,6 @@ set(care_headers
     KeyValueSorter_inst.h
     local_host_device_ptr.h
     local_ptr.h
-    LoopFuser.h
```
robinson96, line 41: these should always be included.  LoopFuser.h provides implementations for the loop macros that are part of CAREs API for all builds for all configurations.

src/care/CMakeLists.txt
```diff
@@ -57,11 +56,15 @@ set(care_headers
 set(care_sources
     care.cpp
     CHAICallback.cpp
-    LoopFuser.cpp
```
robinson96, line 60: I would keep this as well, have it compile to empty (and allow it to be extended in the event LoopFuser disabled still needs CPU implementation for a stripped down LoopFuser class.

test/Benchmarks.cpp
robinson96: There's a broader conversation here - I think it might be reasonable for care::initialize_pool to contain the preprocesor logic to convert things that aren't enabled into fallback options, but I could be convinced otherwise.

src/care/CMakeLists.txt
dtaller: @robinson96 @liu15  Oh no, I meant to close this pull request a long time ago. I think a long time ago I tried to build with spack with the +loop_fuser on, and there was an incompatibility in loop fuser due to conflicting versions. We ended up just updating the versions instead of removing loop fuser from the build system if I recall, this was not needed

dtaller: @adayton1 Don't merge! I meant to close this pull request a long time ago. I think a long time ago I tried to build with spack with the +loop_fuser on, and there was an incompatibility in loop fuser due to conflicting versions. We ended up just updating the versions instead of removing loop fuser from the build system if I recall, this was not needed

dtaller:
Oh no, I meant to close this pull request a long time ago. I think a long time ago I tried to build with spack with the +loop_fuser on, and there was an incompatibility in loop fuser due to conflicting versions which caused loop fuser to fail to compile even when it was turned off. We ended up just updating the versions instead of removing loop fuser from the build system if I recall, this was not needed. Don't merge!!!

test/LoopFuserTest.cxx
```diff
@@ -318,6 +318,10 @@ FUSIBLE_DEVICE bool printAndAssign(care::host_device_ptr<int> B, int i) {
    return B[i] == 1;
 }
 
+#if defined(CARE_GPUCC)
+// parallel scans cannot work on CPU-only builds. This will call  void LoopFuser<REGISTER_COUNT,XARGS...>::flush_parallel_scans,
```
robinson96, line 322: FUSIBLE_LOOP_SCAN macro needs to work and has worked for CPU builds, for loop fuser enabled and not enabled. I suspect this is broken because you removed LoopFuser.h from the install list.

robinson96:
> 
> 
> Oh no, I meant to close this pull request a long time ago. I think a long time ago I tried to build with spack with the +loop_fuser on, and there was an incompatibility in loop fuser due to conflicting versions which caused loop fuser to fail to compile even when it was turned off. We ended up just updating the versions instead of removing loop fuser from the build system if I recall, this was not needed. Don't merge!!!

Phew! I'll stop my review ripping it to shreds. ;).

PR #124, opened 2021-03-19
Update third party libraries

adayton1:
* Updates BLT to the tip of develop past v0.3.6 (required for the latest Umpire)
* Updates CHAI to the tip of develop past v2.3.0
* Updates Umpire to the tip of develop past v5.0.0
* Updates RAJA to v0.13.0
* Removes some code that is defined in the new version of Umpire
* Fixes a build warning

### DIFF
--- a/src/care/LoopFuser.cpp
+++ b/src/care/LoopFuser.cpp
@@ -14,21 +14,6 @@
 #include "umpire/Allocator.hpp"
 #include "umpire/TypedAllocator.hpp"
 
-// TODO: when umpire is updated, you should get duplicate defines of this. Feel free to delete
-// with a vengeance.
-template <typename T, typename U>
-bool operator==(umpire::TypedAllocator<T> const&, umpire::TypedAllocator<U> const&)
-{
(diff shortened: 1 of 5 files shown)


liu15 (approved):

src/care/LoopFuser.cpp
```diff
@@ -14,21 +14,6 @@
 #include "umpire/Allocator.hpp"
 #include "umpire/TypedAllocator.hpp"
 
-// TODO: when umpire is updated, you should get duplicate defines of this. Feel free to delete
-// with a vengeance.
-template <typename T, typename U>
```
dtaller, line 19: Thanks for getting rid of this! This should resolve the spack HIP build conflict that we were seeing.

dtaller (approved):
Thanks, looks good!

src/care/LoopFuser.cpp
adayton1: You're welcome. Thanks for figuring out this was an issue :)

PR #128, opened 2021-05-06
Adds option for array bounds checking for host_device_ptr on the host

adayton1:


src/care/host_device_ptr.h
```diff
@@ -196,6 +196,22 @@ namespace care {
          return *reinterpret_cast<host_device_ptr<const T> const *> (this);
       }
 
+#if defined(CARE_ENABLE_BOUNDS_CHECKING)
+      template <class Index>
+      inline void boundsCheck(const Index i) const {
+         if (i < 0 || i >= (Index) MA::size()) {
```
liu15, line 202: what happens if the array is created as one type by accessed as another (allocated a char but accessed as an int)?

adayton1: Good point. Now using the size in bytes divided by the size of the type that it is when this operator is called.

liu15 (commented):
Al will love this!  Any chance something like this.

Any chance something like this would ever be upstreamed into CHAI?

adayton1:
So, there is a branch in CHAI where this was started, but it stalled because it wasn't obvious how to get this to work on the GPU.

adayton1:
Yeah, Al was the one who asked me for this :)

dtaller (approved):
Looks great! Thanks! Just one question about how this is used in pick() vs operator[], besides that it is all good

src/care/host_device_ptr.h
```diff
@@ -206,11 +222,19 @@ namespace care {
 #if !CARE_LEGACY_COMPATIBILITY_MODE
       const
 #endif
-      { return MA::operator[](i); }
+      {
+#if !defined(CARE_DEVICE_COMPILE) && defined(CARE_ENABLE_BOUNDS_CHECKING)
+         boundsCheck(i);
+#endif
+         return MA::operator[](i);
```
dtaller, line 229: Does this need to be added to the pick() function separately?

src/care/host_device_ptr.h
```diff
@@ -206,11 +222,19 @@ namespace care {
...
+      }
 
 #if CARE_LEGACY_COMPATIBILITY_MODE
      template<typename Idx>
      inline CARE_HOST_DEVICE T& operator[](const Idx i) const {
+#if !defined(CARE_DEVICE_COMPILE) && defined(CARE_ENABLE_BOUNDS_CHECKING)
+        boundsCheck(i);
```
dtaller, line 236: Does this need to be added to the pick() function separately?

dtaller (commented):
Do these need to be added to the pick() function separately, rather than just operator[]?

src/care/host_device_ptr.h
adayton1: Done.

PR #137, opened 2021-06-08
Update to BLT v0.4.0

adayton1:


dtaller (approved):
Looks good to me. I assume you are updating this for HIP related fixes?

dtaller:
Actually it  looks like tests failed, I should have looked more closely . I'll look at this again when that is sorted out.

adayton1:
The tests actually failed due to the default environment on GitHub servers being updated (looks like a newer cmake caused the problem). Ben just put up a pull request to fix that, so we should be good to go.

adayton1:
Yeah, BLT v0.4.0 has better support for HIP.

dtaller:
> Yeah, BLT v0.4.0 has better support for HIP.

Yes, previously we made incremental hip related changes to blt and built care with blt develop. If this fixed version has all the HIP fixes we need, it will make development easier. Thanks!

liu15 (approved):

PR #148, opened 2021-08-16
Fix host_ptr to local_ptr conversion when implicit conversions are disabled

adayton1:


liu15 (approved):
What was the previous behavior?

dtaller (approved):

adayton1:
If you tried to pass a host_ptr to a function that accepts a local_ptr (which should be valid), there was a compile error. This used to work before we disabled implicit conversions. Now it should work in both cases where implicit conversions are enabled AND where they are disabled.

agcapps (approved):
Thanks much, @adayton1 .

PR #153, opened 2021-09-01
Fix the loop fuser tests on the host

adayton1:
- Add toss clang host config
- Fix build warnings (e.g. unused variables, deprecated behavior for implicitly defined copy constructors)
- Rename LoopFuserTest.cxx to TestLoopFuser.cxx to be consistent with all other test file names
- Fix memory leaks in loop fuser tests
- Fix loop fuser tests on the host


liu15 (approved):

test/TestLoopFuser.cxx
```diff
@@ -143,7 +135,8 @@ GPU_TEST(TestPacker, packFixedRangeMacro) {
 
    care::gpuDeviceSynchronize(__FILE__, __LINE__);
 
-   // pack should  not have happened yet so
+#ifdef CARE_GPUCC
```
dtaller, line 138: Maybe add comment by CARE_GPUCC protection is needed here

dtaller: Maybe add comment by CARE_GPUCC protection is needed here and also below in other places where you did this, just for clarity

dtaller (approved):
In the Loop Fuser Test file, just add a comment on why #ifdef CARE_GPUCC was needed in the places you added it, just for clarity.

besides that, looks good!

robinson96 (approved):
LGTM

test/TestLoopFuser.cxx
adayton1: Updated the comment (now it specifically mentions the device, so I think that should be sufficient).

PR #156, opened 2021-09-08
Add option for looping backwards over parallel loops

adayton1:
* This may be useful for locating race conditions in a host build

### DIFF
--- a/src/care/forall.h
+++ b/src/care/forall.h
@@ -80,14 +80,10 @@ namespace care {
 #endif
 
 #if CARE_ENABLE_PARALLEL_LOOP_BACKWARDS
-         RAJA::RangeStrideSegment rangeSegment(start, end, 1);
-
-         if (s_reverseLoopOrder) {
-            rangeSegment = RAJA::RangeStrideSegment(end - 1, start - 1, -1);
-         }
-         else {
-            rangeSegment = RAJA::RangeStrideSegment(start, end, 1);
-         }
+         RAJA::RangeStrideSegment rangeSegment =
+            s_reverseLoopOrder ?
+            RAJA::RangeStrideSegment(end - 1, start - 1, -1) :
+            RAJA::RangeStrideSegment(start, end, 1);
 #else
          RAJA::RangeSegment rangeSegment = RAJA::RangeSegment(start, end);
 #endif
--- a/src/care/forall.h
+++ b/src/care/forall.h
@@ -77,7 +77,7 @@ namespace care {
 #endif
 
 #if CARE_ENABLE_PARALLEL_LOOP_BACKWARDS
-         RAJA::RangeStrideSegment rangeSegment;
+         RAJA::RangeStrideSegment rangeSegment(start, end, 1);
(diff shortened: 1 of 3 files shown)


src/care/forall.h
```diff
@@ -71,14 +71,33 @@ namespace care {
...
+         RAJA::RangeStrideSegment rangeSegment;
+
+         if (std::is_same<ExecutionPolicy, openmp>::value ||
+             std::is_same<ExecutionPolicy, gpu>::value ||
+             std::is_same<ExecutionPolicy, parallel>::value ||
+             std::is_same<ExecutionPolicy, gpu_simulation>::value) {
+            rangeSegment = RAJA::RangeStrideSegment(end - 1, start - 1, -1);
+         }
+         else {
+            rangeSgement = RAJA::RangeStrideSegment(start, end, 1);
```
liu15, line 89: Sgement -> Segment

src/care/forall.h
```diff
@@ -71,14 +71,33 @@ namespace care {
       const int length = end - start;
 
       if (length != 0) {
+
 #ifndef CARE_DISABLE_RAJAPLUGIN
          RAJAPlugin::pre_forall_hook(ExecutionPolicyToSpace<ExecutionPolicy>::value, fileName, lineNumber);
 #endif
+
+#if CARE_ENABLE_PARALLEL_LOOP_BACKWARDS
+         RAJA::RangeStrideSegment rangeSegment(start, end, 1);
```
liu15, line 80: No default constructor, eh?

Would it be better either define segmentStart, segmentEnd, and segmentStride as variables can call rangeSegment(segmentStart, segmentEnd, segmentStride)?  Or defined a boolean isParallel and initialize using a ternary operator?

adayton1: I'll go with the ternary operator.

src/care/forall.h
adayton1: Fixed.

adayton1:
OK, now this is working and ready for review.

src/care/forall.h
```diff
@@ -193,6 +228,10 @@ namespace care {
 #else
```
dtaller, line 228: This is totally unrelated to this PR, but I noticed we have a few lines above

forall(RAJA::hip_exec<CARE_CUDA_BLOCK_SIZE, CARE_CUDA_ASYNC>{},
             fileName, lineNumber, start, end, body);

above. In the future we should remember to either have CARE_GPU_BLOCK_SIZE (just rename it) or have separate CARE_CUDA_BLOCK_SIZE and CARE_HIP_BLOCK_SIZE

dtaller (approved):
Looks good to me

src/care/forall.h
adayton1: Yeah, we'll need to clean that up. Thanks for pointing that out.

adayton1: I created an issue so that we remember to fix this.

liu15 (approved):

PR #157, opened 2021-09-08
Best practices with BLT

adayton1:
* Remove uses of blt_register_library
* Use blt_import_library where a native CMake target is not available
* Use native CMake target where available


liu15 (approved):

dtaller:
Looks good to me!

dtaller (approved):

PR #158, opened 2021-09-22
Fix the loop fuser tests for CUDA builds

adayton1:
* Enable the RAJA plugin in CHAI, which is needed for the loop fuser to work
* Add a temporary workaround for a test case that causes other test failures


liu15 (approved):

dtaller (approved):

robinson96 (approved):
LGTM

adayton1:
The CI timed out. I'm looking into how to extend the allocation time.

PR #159, opened 2021-09-23
Fix HIP build

adayton1:
* Fix the rznevada HIP build
* Add a config file for rzwhamo


dtaller (approved):
Looks good to me

liu15 (approved):

robinson96 (approved):
LGTM

adayton1:
The RAJA_ENABLE_HIP_INDIRECT_FUNCTION_CALL option is needed to turn on loop fusion in RAJA for HIP. Loop fusion basically stores up a bunch of function pointers and makes indirect function calls using the function pointers. That is not supported in older versions of rocm.

PR #172, opened 2021-12-20
Flush non phased actions when lengths are exceeded; add possible flus…

liu15:
…h points in a couple more places


liu15:
There is currently a GPU runtime error here.

src/care/LoopFuser.h
```diff
@@ -1254,13 +1254,23 @@ void LoopFuser<REGISTER_COUNT, XARGS...>::registerAction(const char * fileName,
...
 
 #define FUSIBLE_KERNEL_PHASE(PRIORITY) FUSIBLE_KERNEL_PHASE_R(PRIORITY, CARE_DEFAULT_LOOP_FUSER_REGISTER_COUNT)
 
-#define FUSIBLE_PHASE_RESET FusedActionsObserver::getActiveObserver()->reset_phases(__FILE__, __LINE__);
+#define FUSIBLE_PHASE_RESET   if (FusedActions::flush_now) { \
+                                 for ( FusedActions *__fuser__ : { \
+                                    static_cast<FusedActions *> (LOOPFUSER(256)::getInstance()),\
+                                    static_cast<FusedActions *> (LOOPFUSER(128)::getInstance()),\
+                                    static_cast<FusedActions *> (LOOPFUSER(64)::getInstance()),\
+                                    FUSED_ACTION_INSTANCE_32 \
```
adayton1, line 1268: Does `FusedActionsObserver::getActiveObserver()` also need to be accounted for here?

src/care/LoopFuser.h
```diff
@@ -1341,7 +1351,7 @@ void LoopFuser<REGISTER_COUNT, XARGS...>::registerAction(const char * fileName,
                                     FUSED_SCANVAR[__fusible_global_index__] = SCANVAR[INDEX]; \
                                  } \
                                  }, \
-                              2, __fusible_scan_pos__ , SCANVAR); }
+                              2, __fusible_scan_pos__ , SCANVAR); } FUSIBLE_PHASE_RESET
```
adayton1, line 1354: I wonder if we need to think out the design a bit more (or perhaps just the naming). It is not intuitive for me to look at this and try to think about how fusible phases are involved here. I think I would rather have another macro like FUSIBLE_FLUSH_POINT or FUSIBLE_FLUSH_IF_NEEDED even if it does the same thing.

src/care/LoopFuser.h
liu15: FusedActionsObserver::getActiveObserver()->reset_phases(__FILE__, __LINE__);
calls flushActions if flush_now is set.

src/care/LoopFuser.h
```diff
@@ -1254,13 +1254,23 @@ void LoopFuser<REGISTER_COUNT, XARGS...>::registerAction(const char * fileName,
...
 
-#define FUSIBLE_KERNEL_R_END return 0;}); }
+#define FUSIBLE_KERNEL_R_END return 0;}); } FUSIBLE_PHASE_RESET
 #define FUSIBLE_KERNEL_PHASE_R_END FUSIBLE_KERNEL_R_END
 #define FUSIBLE_KERNEL_END FUSIBLE_KERNEL_R_END
 
 #define FUSIBLE_KERNEL_PHASE(PRIORITY) FUSIBLE_KERNEL_PHASE_R(PRIORITY, CARE_DEFAULT_LOOP_FUSER_REGISTER_COUNT)
 
-#define FUSIBLE_PHASE_RESET FusedActionsObserver::getActiveObserver()->reset_phases(__FILE__, __LINE__);
+#define FUSIBLE_PHASE_RESET   if (FusedActions::flush_now) { \
```
adayton1, line 1263: Could we make this a function rather than a macro?

src/care/LoopFuser.h
adayton1: I guess it's probably accounted for in the reset_phases call, but it would be nice for things to be consistent.

src/care/LoopFuser.h
liu15: Yeah, perhaps FUSIBLE_PHASE_RESET could call FUSIBLE_FLUSH_IF_NEEDED?

src/care/LoopFuser.h
liu15: Would that be a static method in FusedActions?  Or FusedActionsObserver?   It would have to be defined later, as it needs to get instances of LoopFuser.

adayton1 (approved):
LGTM

robinson96:
LGTM

robinson96 (approved):
LGTM

PR #177, opened 2022-03-15
Feature/probinso/map improvements

robinson96:
care::host_device_map improvements.
DETAILS:
addition of a size(), clear(), and reserve() methods to API. 
constructor now takes an additional argument - a mapped_type value to be returned if an at(key) call can not find an entry at that key. 
Addition of a CARE_STREAM_MAP_LOOP macro to provide GPU iteration over elements in the host_device_map.


adayton1 (approved):
LGTM

liu15 (approved):

PR #178, opened 2022-03-17
Add GIDTYPE instances

liu15:
This fixes our sierra GID64 build, but is the GIDTYPE key type important enough to instantiate here?  Or do we want even more types instantiated? 

I think only the const host_device_ptr GIDTYPE BinarySearch is needed, but I added the others for symmetry.

### DIFF
--- a/src/care/care_inst.h
+++ b/src/care/care_inst.h
@@ -38,13 +38,17 @@
 #include "care/KeyValueSorter_inst.h"
 
 #if CARE_HAVE_LLNL_GLOBALID
+
 #define CARE_TEMPLATE_KEY_TYPE int
 #define CARE_TEMPLATE_ARRAY_TYPE globalID
 #include "care/KeyValueSorter_inst.h"
 
+#if GLOBALID_IS_64BIT
 #define CARE_TEMPLATE_KEY_TYPE GIDTYPE
 #define CARE_TEMPLATE_ARRAY_TYPE int
 #include "care/KeyValueSorter_inst.h"
+#endif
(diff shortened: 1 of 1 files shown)


adayton1 (approved):
LGTM

src/care/care_inst.h
```diff
@@ -41,6 +41,10 @@
 #define CARE_TEMPLATE_KEY_TYPE int
 #define CARE_TEMPLATE_ARRAY_TYPE globalID
 #include "care/KeyValueSorter_inst.h"
+
```
robinson96, line 44: Don't you need to protect this for whether GID64 is enabled? Otherwise GIDTYPE resolves to int and you are defining <int,int> twice.

liu15: The endif on line 48 is for #if CARE_HAVE_LLNL_GLOBALID

liu15:
I am going to test this to see if I can remove the BinarySearch instantiations now that I have the KeyValue instances.

src/care/care_inst.h
adayton1: I think Peter may be right that the non-gid64 variant is going to break with this change (CARE_HAVE_LLNL_GLOBALID is defined in both cases, just GIDTYPE changes between int and long). If this does break, you could just instantiate both int and long.

liu15: Oh, I get it.  I can add the protection.

agcapps (approved):
Thanks, @liu15 .

PR #179, opened 2022-03-23
host_device_map fixes: add GPU_SIM mode, add bounds checking for empl…

liu15:
host_device_map fixes:
- add GPU_SIM mode
- add bounds checking for emplace
- reserve only increases size

KeyValueSorter fixes
- GPU_SIM mode added

### DIFF
--- a/src/care/host_device_map.h
+++ b/src/care/host_device_map.h
@@ -70,6 +70,7 @@ namespace care {
         // emplace a key value pair
         inline void emplace(key_type key, mapped_type val) const {
            m_map->emplace(key, val);
+           // TODO Add control for this check
            if (m_map->size() > m_max_size) {
               printf("[CARE] Warning: host_device_map exceeds max size %d > %d\n", (int)m_map->size(), m_max_size);
            }
@@ -145,6 +146,7 @@ namespace care {
         inline CARE_DEVICE void emplace(key_type key, mapped_type val) const {
            care::local_ptr<int> size_ptr = m_size_ptr;
            int index = ATOMIC_ADD(size_ptr[0], 1);
+           // TODO Add control for this check
            if (size_ptr[0] > m_max_size) {
               printf("[CARE] Warning: host_device_map exceeds max size %d > %d\n", size_ptr[0], m_max_size);
            }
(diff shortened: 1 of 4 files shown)


robinson96 (approved):
LGTM

src/care/host_device_map.h
```diff
@@ -69,6 +70,9 @@ namespace care {
         // emplace a key value pair
         inline void emplace(key_type key, mapped_type val) const {
            m_map->emplace(key, val);
+           if (m_map->size() > m_max_size) {
```
adayton1, line 73: Could this check be guarded with a macro (e.g. CARE_DEBUG or CARE_ENABLE_BOUNDS_CHECKING)?

src/care/host_device_map.h
```diff
@@ -136,7 +143,11 @@ namespace care {
 
         // emplace a key value pair, using return of atomic increment to provide the initial insertion index
         inline CARE_DEVICE void emplace(key_type key, mapped_type val) const {
-           int index = ATOMIC_ADD(m_size_ptr[0], 1);
+           care::local_ptr<int> size_ptr = m_size_ptr;
+           int index = ATOMIC_ADD(size_ptr[0], 1);
+           if (size_ptr[0] > m_max_size) {
```
adayton1, line 148: Same here.

src/care/host_device_map.h
liu15: I'm open to whatever you guys prefer.

src/care/host_device_map.h
liu15: But I think we would definitely want this active in our default builds.

adayton1: Oof, to have it active for debug builds but turn it off for release builds is surprisingly difficult (especially with explicit instantiations and public packages).

adayton1: I'm just worried that since emplace is often called in a loop, the overhead of bounds checking will be significant.

adayton1 (approved):
LGTM

liu15:
weird.  Does this require @robinson96  to reapprove?

PR #184, opened 2022-04-07
Add care::copy_n method

adayton1:
Attempting to match std::copy_n


src/care/algorithm_impl.h
```diff
@@ -979,6 +979,20 @@ CARE_INLINE void fill_n(care::host_device_ptr<T> arr, Size n, const U& val)
...
+ * Author(s) : Alan Dayton
+ * Purpose   : Copies one ManagedArray into another.
+ * ************************************************************************/
+template <class T, class Size, class U>
+CARE_INLINE void copy_n(care::host_device_ptr<T> in, Size n,
+                        care::host_device_ptr<U> out)
+{
+   CARE_STREAM_LOOP(i, 0, n) {
+      out[i] = in[i];
+   } CARE_STREAM_LOOP_END
```
liu15, line 993: Since this is inline, we don't need instantiations, right?

liu15 (approved):

src/care/algorithm_impl.h
```diff
@@ -979,6 +979,20 @@ CARE_INLINE void fill_n(care::host_device_ptr<T> arr, Size n, const U& val)
    } CARE_STREAM_LOOP_END
 }
 
+/************************************************************************
+ * Function  : copy_n
+ * Author(s) : Alan Dayton
+ * Purpose   : Copies one ManagedArray into another.
+ * ************************************************************************/
+template <class T, class Size, class U>
+CARE_INLINE void copy_n(care::host_device_ptr<T> in, Size n,
```
dtaller, line 988: Can the in array be made const?

dtaller: Or on the other hand I don't think std::copy_n is const , maybe that change isn't needed?

dtaller (approved):

src/care/algorithm_impl.h
liu15: Const would be nice, if not in the signature, at least before the lambda capture.

src/care/algorithm_impl.h
adayton1: I believe that is the case.

src/care/algorithm_impl.h
adayton1: Just made it const.

adayton1: Looks like I need both a const and non-const overload (I just made the non-const overload call the const overload).

adayton1:
@dtaller, @liu15 I finally addressed your review comments from a long time ago. This is is ready for another review pass.

dtaller:
Looks good to me! Thanks!

liu15:
Looks good to me as well.

PR #187, opened 2022-06-03
Fix bounds checking when RM disabled

liu15:


src/care/host_device_ptr.h
```diff
@@ -199,7 +199,11 @@ namespace care {
 #if defined(CARE_ENABLE_BOUNDS_CHECKING)
       template <class Index>
       inline void boundsCheck(const Index i) const {
-         if (i < 0 || i >= (Index) (MA::m_pointer_record->m_size / sizeof(T))) {
+         if (i < 0
+#if !defined (CHAI_DISABLE_RM)
```
dtaller, line 203: I wonder if we should add a warning message or explicit documentation somewhere, to let the user know that enabling bounds checking while disabling rm means that we cannot check the upper bound. 

I just worry that a user might be surprised that the upper bound is not being checked even when they set CARE_ENABLE_BOUNDS_CHECKING on.

dtaller (approved):
Approved, but a warning or documentation as explained in my comment would be good

src/care/host_device_ptr.h
dtaller: Or even require that if CARE_ENABLE_BOUNDS_CHECKING is on, that CHAI_DISABLE_RM be off (but maybe that is too restrictive?)

liu15: Our standard toss3 build does not have the RM, so disabling it would be too restrictive.  I think a run-time warning would be too noisy.  Where would be the best place to put the documentation?

liu15: This fix is from Al.

adayton1 (approved):
LGTM

PR #188, opened 2022-06-06
Add care::min and care::max, remove CARE_MIN and CARE_MAX

liu15:


src/care/algorithm_decl.h
```diff
@@ -21,10 +21,20 @@
 #include "LLNL_GlobalID.h"
 #endif // CARE_HAVE_LLNL_GLOBALID
 
-#define CARE_MAX(a,b) a > b ? a : b
-#define CARE_MIN(a,b) a < b ? a : b
-
 namespace care {
+
+template <typename T>
+CARE_HOST_DEVICE CARE_INLINE T max(const T a, const T b)
```
adayton1, line 27: Can you use references to match the signature of std::max?

```
template<class T> 
const T& max(const T& a, const T& b)
```

adayton1: I guess that could prohibit using literals, though, so maybe not.

adayton1 (approved):
LGTM

src/care/algorithm_decl.h
robinson96: whoa I had no idea max returned a reference. But makes sense when you start to think about it from a performance / avoiding copy construction perspective.

robinson96: https://en.cppreference.com/w/cpp/algorithm/max

They have examples with literals - I'm not sure if those work with the templated &  signature and the inlining makes it all happy or if it ends up going through the initializer list signature when you use literals.

robinson96: I think we should try matching the standard signature.

robinson96: https://en.cppreference.com/w/cpp/language/reference_initialization
I think this explains why this works

src/care/algorithm_decl.h
```diff
@@ -21,10 +21,26 @@
 #include "LLNL_GlobalID.h"
 #endif // CARE_HAVE_LLNL_GLOBALID
 
-#define CARE_MAX(a,b) a > b ? a : b
-#define CARE_MIN(a,b) a < b ? a : b
-
 namespace care {
+
+template <typename T>
+CARE_HOST_DEVICE CARE_INLINE T abs(const T a)
```
adayton1, line 27: Should this also use references?

adayton1:
Could you add some simple test cases for these functions?

src/care/algorithm_decl.h
liu15: The std::abs does not use references, so I copied that.  The behavior is different since this is templated, so I could go either way.

robinson96 (approved):

src/care/algorithm_decl.h
liu15: Let me know if you guys prefer a reference for the argument

adayton1: Don't worry about it for now.

PR #189, opened 2022-06-07
Fixes for loop fuser

liu15:
Fix FUSIBLE_KERNEL_PHASE


robinson96 (approved):
LGTM

src/care/LoopFuser.h
```diff
@@ -1254,6 +1255,7 @@ void LoopFuser<REGISTER_COUNT, XARGS...>::registerAction(const char * fileName,
 
 #define FUSIBLE_KERNEL_R_END return 0;}); } FUSIBLE_FLUSH_IF_NEEDED
 #define FUSIBLE_KERNEL_PHASE_R_END FUSIBLE_KERNEL_R_END
+#define FUSIBLE_KERNEL_PHASE_END FUSIBLE_KERNEL_R_END
```
liu15, line 1258: Should this have the FUSIBLE_FLUSH_IF_NEEDED or not?

liu15: Also, I noticed that FUSIBLE_LOOP_STREAM_R doesn't have if (END > START), but FUSIBLE_LOOP_PHASE_R does.  Is that intentional?

robinson96: > Should this have the FUSIBLE_FLUSH_IF_NEEDED or not?

I think it is good as is (calling FUISBLE_FLUSH_IF_NEEDED through FUSIBLE_KERNEL_R_END)

robinson96: It does seem like the END > START check is worth doing for all the loop types at first glance. It's been a while since I did this, but I don't think I bumped into anything that required calling the infrastructure for what would be no-ops.

liu15: How about the FUSIBLE_FLUSH_IF_NEEDED?  It is omitted in FUSIBLE_LOOP_PHASE_R_END

adayton1 (approved):
LGTM

PR #190, opened 2022-06-08
Remove conditional dependence on GPU_ACTIVE

adayton1:
When care headers were included in some compilation units with GPU_ACTIVE defined and others without it defined, the linker got to pick which definition of some functions to use (in the case where a STREAM LOOP was used in a header file). This pull request removes any dependence on GPU_ACTIVE.


adayton1:
I plan to do similar work to remove OPENMP_ACTIVE in the future, but punted on that for the time being since it was causing problems with clang-query.

liu15 (approved):

robinson96 (approved):
LGTM

PR #192, opened 2022-06-09
Add explicit instantiations for care::copy_n

adayton1:


dtaller (approved):

dtaller:
Did these changes occur after some issue when incorporating things into the client code?

robinson96 (approved):
LGTM

adayton1:
> Did these changes occur after some issue when incorporating things into the client code?

Yes, a client code ran into issues since they configure CARE to use explicit instantiations. At some point it might be worth considering using precompiled headers rather than explicit instantiations. I feel like I would rather have the headache in the build system than in the actual source code.

PR #197, opened 2022-06-30
Host device map fixes: implement host iterator_at and unify map loops…

liu15:
…; reset size in host clear(); fix loop fuser map macros for host; remove START argument for MAP macros


src/care/host_device_map.h
```diff
@@ -90,34 +94,52 @@ namespace care {
...
+              }
+           }
+           ++(*m_next_iterator_index);
+           return (*m_iterator)++;
+        }
 
       private:
          // we do a heap allocated map to ensure no deep copies occur during lambda capture
          std::map<key_type, mapped_type> * m_map = nullptr;
+         mutable typename std::map<key_type, mapped_type>::iterator * m_iterator = nullptr;
```
robinson96, line 141: you shouldn't need the mutables since the pointer value itself isn't changing in a lambda context

liu15: Ah, yes.  I think I was missing a dereference in an earlier version (which caused the compiler to complain about const-ness).

robinson96 (approved):
LGTM

adayton1 (approved):
LGTM

PR #201, opened 2022-08-25
Use perfect forwarding for lambdas

adayton1:
Avoid unnecessary copies of temporary lambdas passed to RAJA.
Also remove unnecessary macros.


dtaller:
Looks good to me! Thanks for the update

dtaller (approved):

dtaller:
When you push this update into client codes let me know if / how much improvement you see from the perfect forwarding, curious to see.

robinson96 (approved):
LGTM. Have we measured a performance improvement?

agcapps (approved):
Thanks, Alan.

liu15 (approved):

adayton1:
There seems to be a slight (1.01x) improvement in some application testing. I don't have benchmarks for it in CARE at the moment, though.

PR #202, opened 2022-10-12
Add check for stale host data view

adayton1:
* To build with new check, set the following CMake variable: CARE_ENABLE_STALE_DATA_CHECK=ON
* Will abort with an error message if stale data is detected.


agcapps (approved):
Thanks @adayton1 !

src/care/host_ptr.h
```diff
@@ -17,6 +17,11 @@
 // Std library headers
 #include <cstddef>
 
+#if defined(CARE_ENABLE_STALE_DATA_CHECK) && !defined(CHAI_DISABLE_RM) && (defined(CHAI_ENABLE_CUDA) || defined(CHAI_ENABLE_HIP) || defined(CHAI_ENABLE_GPU_SIMULATION_MODE))
```
dtaller, line 20: This is optional, but you could make a variable defined as defined(CARE_ENABLE_STALE_DATA_CHECK) && !defined(CHAI_DISABLE_RM) && (defined(CHAI_ENABLE_CUDA) || defined(CHAI_ENABLE_HIP) || defined(CHAI_ENABLE_GPU_SIMULATION_MODE)) since it is long and you use it twice

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ option(CARE_NEVER_USE_RAJA_PARALLEL_SCAN "Disable RAJA parallel scans in SCAN lo
 #  practice for Release / RelWithDebInfo configurations.
 option(CARE_ENABLE_FUSER_BIN_32 "Enable the 32 register fusible loop bin." OFF)
 option(CARE_ENABLE_PARALLEL_LOOP_BACKWARDS "Reverse the start and end for parallel loops." OFF)
+option(CARE_ENABLE_STALE_DATA_CHECK "Enable checking for stale data." OFF)
```
dtaller, line 45: This is optional, but you could add a comment here saying that this option is only applicable for GPU or GPU simulation modes. 

On the other hand, that might be implied given that if you are only on the host you never transfer data so stale data would not be relevant. Up to you.

dtaller (approved):
Thanks for adding this!

CMakeLists.txt
adayton1: Good idea! Done.

src/care/host_ptr.h
adayton1: I'm not going to worry about this since it is already in one application's fork of CARE.

PR #208, opened 2022-11-30
Feature/probinso/race detect

robinson96:
Adds an Accessor pattern to the host_device_ptr, enables an accessor that, when CARE_ENABLE_RACE_DETECTION is on, records which threads access each index and performs a check for thread collisions where values are changed within a parallel loop context.


liu15 (approved):

adayton1 (approved):
LGTM

adayton1:
@robinson96, looks like the tests failed because CARE is still on C++11 by default and you used std::remove_const_t which is in C++14. Can you bump the default to C++14 (I think it's BLT_CXX_STD in CMakeLists.txt)?

PR #212, opened 2023-01-21
GitHub CI fixes

adayton1:
Our GitHub CI was relying on the default version of CMake, which was recently advanced to CMake 3.25. Unfortunately, that version has a bug where check_cxx_source_compiles does not respect the c++ standard, resulting in a feature of c++17 erroneously being used in Umpire.


adayton1:
The default gcc also advanced, causing a failure when BLT builds gbenchmark. There is a patch in BLT v0.5.0.

adayton1:
I also am removing the gcc 4.9.3 build from our gitlab CI. It is very old and does not fully support c++14, which is what CARE should be on.

agcapps:
Who is jwlawson?  Is that user or the CMake setter-upper he or she provides ever going to go away?  We may want to get a local copy of that script.

liu15 (approved):

agcapps (approved):
If this fixes CARE's GitHub CI, that's what we need.  But I'm still curious about the possibility of the fix (jwlawson?) going away.  (Maybe that concern isn't warranted.)

adayton1:
Github has a "Marketplace" of CI actions that people can share and reuse. I don't believe this action will go away. If it does, then we can find another action, write our own, or just drop github testing and only test on our systems.

adayton1:
This is where the Github action comes from: https://github.com/marketplace/actions/actions-setup-cmake. Here is an alternative action: https://github.com/marketplace/actions/get-cmake

adayton1:
If it was just a single file, we might be able to take a local copy, but it is actually a small repo and I'm not sure which files would be needed. But these "actions" are intended to be stable.

PR #216, opened 2023-02-09
Accessor fixes for GPU builds, use blt::hip instead of hip target

robinson96:
make lop fuser flush length configurable.


adayton1 (approved):
LGTM

src/care/LoopFuser.cpp
```diff
@@ -25,8 +25,8 @@
 CARE_DLL_API int FusedActions::non_scan_store = 0;
 CARE_DLL_API bool FusedActions::verbose = false;
 CARE_DLL_API bool FusedActions::very_verbose = false;
-// set flush length to 8M by default
-CARE_DLL_API int FusedActions::flush_length = 8388608; 
+// set flush length to 8M by default (default value is defined in imports/care/CMakeLists.txt)
```
liu15, line 28: This should probably not include imports/care in the path.

liu15 (approved):
Just the minor comment.

PR #218, opened 2023-03-14
Update dependencies

adayton1:
CAMP: v2022.10.1
Umpire: v2022.10.0
RAJA: v2022.10.5
CHAI: v2022.10.0

Also update CMake variable names when building with submodules (CHAI/RAJA/Umpire now prefix a lot of CMake variables, which is good, but requires us to make some changes).

Spack packages for the submodules will be updated in a follow on pull request. It's too complicated to try to do all at once (I've tried and failed multiple times). This at least gets us closer.


dtaller (approved):

agcapps (approved):
Thumbs up

PR #219, opened 2023-03-23
Standalone CARE build fixes on rzansel

adayton1:
* Use the cub target imported by RAJA if available
* Update the blueos host config
* Remove ambiguous operator== overload in host_device_ptr and just rely on the ones for chai::ManagedArray


src/care/host_device_ptr.h
```diff
@@ -456,14 +456,6 @@ namespace care {
       CARE_HOST void move(ExecutionSpace space) {
          MA::move(chai::ExecutionSpace((int) space));
       }
-      
-      CARE_HOST_DEVICE inline bool operator ==(host_device_ptr<T, Accessor> const & right) const {
```
adayton1, line 460: Alternatively, I could move this outside the class and add nullptr overloads.

adayton1: The CHAI version only compares the active pointers, but I think we enforce consistency enough that that should be sufficient. And most often we just compare against nullptr.

adayton1: To be honest, I'm not convinced either ours or CHAI's implementation is really correct. Perhaps we ought to be comparing the pointer records.

dtaller: I'll approve this MR in any case, but yes something to discuss with in the future

dtaller (approved):

liu15 (approved):

agcapps (approved):
Thanks, @adayton1 .

PR #220, opened 2023-03-23
Add interface for creating/using array views

adayton1:


configs/lc/toss/clang.cmake
```diff
@@ -1,3 +1,7 @@
-set(CMAKE_C_COMPILER "/usr/tce/packages/clang/clang-11.0.1/bin/clang" CACHE FILEPATH "Path to C compiler")
```
agcapps, line 1: No CMAKE_C_COMPILER anymore?

src/care/ArrayView.h
```diff
@@ -0,0 +1,68 @@
...
+   ///
+   /// Creates a 2D view on top of a flat allocation (non-owning semantics)
+   ///
+   /// Moves the data to the default execution space (determined at
+   /// configuration time) and wraps it in a RAJA::View. When using with
+   /// a RAJA loop, extent2 should correspond to the length of the RAJA
+   /// loop for optimal performance.
+   ///
+   /// @param[in] data Container for the allocation
+   /// @param[in] extent1 Length of the 1st dimension
```
liu15, line 33: does there need to some description of what 1st and 2nd dimension are?

configs/lc/toss/clang.cmake
adayton1: CARE just uses C++, but I just realized that Umpire has some C code, and this config is used to drive building the dependencies as well as CARE in some cases. So I added the C compiler back in.

src/care/ArrayView.h
adayton1: I'm not sure what that description should be. I added the comment starting at line 27 to try to describe which length should go last in the argument list. If you have suggestions, I'm certainly open to providing more clarification.

liu15: Yeah, I don't know either.

agcapps:
> Looks good. Just curious if you meant to leave out the C compiler.

Thanks for adding that back in.

src/care/ArrayView.h
adayton1: I'm going to go ahead with merging this, but I'm happy to revisit when we start using this in applications and find that more clarification is needed.

PR #221, opened 2023-04-11
Improvements to ArrayView interface

adayton1:


src/care/ArrayView.h
```diff
@@ -6,6 +6,23 @@
 #include "RAJA/util/View.hpp"
 
 namespace care {
+   template <class T>
+   using ArrayView1D = RAJA::View<T, RAJA::Layout<1>>;
+
+   template <class T>
+   using ArrayCView1D = ArrayView1D<const T>;
```
dtaller, line 13: This is nit picky, but why did you call this `ArrayCView1D `rather than `ArrayViewC1D` (Array View Constant 1D) or `CArrayView1D` (Constant Array View 1D)?

dtaller: I guess elsewhere you call this a const view so maybe that makes sense nvm

dtaller (approved):
Looks good to me

src/care/ArrayView.h
adayton1: I was actually going to ask what you think of the naming. Would it make sense to drop "Array" from the naming scheme and just have View1D, ConstView1D, View2D, ConstView2D, etc...?

liu15 (approved):

src/care/ArrayView.h
dtaller: That makes sense. We have RAJA::View, so having care::View would be consistent and it would be clear as long as it is used in a care scope. Spelling out Constant makes things clear also, harder to miss the difference. But I would ask some of the others before making a sweeping change.

adayton1: This isn't used anywhere yet, so now is a great time to make changes.

PR #224, opened 2023-06-22
Use RAJA plugin capability

neelakausik:
CARE had custom infrastructure for doing things before and after RAJA loops. This branch moves debug and profile related code into actual RAJA plugins and makes both plugins optional. It also introduces a requirement to use CHAI's RAJA plugin to make sure data is moved to the correct execution space.


tpl/cub
adayton1, line 1: Did you mean to change cub?

src/care/forall.h
```diff
@@ -120,6 +120,20 @@ namespace care {
       forall(RAJA::seq_exec{}, fileName, lineNumber, start, end, std::forward<LB>(body));
    }
 
+   ////////////////////////////////////////////////////////////////////////////////
+   template <typename LB>
```
adayton1, line 124: The stream changes should be in a different branch.

src/care/RAJAPlugin.cpp
```diff
@@ -25,6 +25,8 @@
 
 #include <unordered_set>
 
+using namespace std;
```
adayton1, line 28: This doesn't look used.

.radiuss-ci
adayton1, line 1: I doubt you meant to change this.

benchmarks/BenchmarkPlugin.cpp
```diff
@@ -0,0 +1,57 @@
+//////////////////////////////////////////////////////////////////////////////////////
```
adayton1, line 1: I think we can remove this file since BenchmarkForall.cpp already covers the same functionality (albeit through the macro layer).

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,46 @@
+#ifndef _CARE_ProfilePlugin_H_
+#define _CARE_ProfilePlugin_H_
+
+#include "RAJA/util/PluginStrategy.hpp"
+#include "chai/ExecutionSpaces.hpp"
+#include "care/config.h"
+#include "care/CHAICallback.h"
+#include "care/RAJAPlugin.h"
+
+using namespace care;
```
adayton1, line 10: "using namespace" is a really bad idea in a header file. It can create function name and class name conflicts for anything that uses CARE.

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace care{
+
+class RAJAPlugin;
```
adayton1, line 14: Why are you forward declaring this class?

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,46 @@
...
+
+}
+
+namespace chai{
```
adayton1, line 18: This should be in the care namespace.

src/care/forall.h
adayton1: To undo all the changes in a file you can do "git checkout develop -- <filename>".

For example, "git checkout develop -- src/care/forall.h"

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+// Std library headers
+#if defined(CARE_DEBUG) && !defined(_WIN32)
+#include <execinfo.h>
+#endif // defined(CARE_DEBUG) && !defined(_WIN32)
+
+#include <unordered_set>
+
+using namespace care;
+
+namespace chai{
```
adayton1, line 25: Use the care namespace rather than chai.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+#endif
+
+// Std library headers
+#if defined(CARE_DEBUG) && !defined(_WIN32)
+#include <execinfo.h>
+#endif // defined(CARE_DEBUG) && !defined(_WIN32)
+
+#include <unordered_set>
+
+using namespace care;
```
adayton1, line 23: Technically not wrong, but doesn't match the rest of CARE stylistically. We don't really do "using namespace" anywhere.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+#if defined(__CUDACC__) && CARE_HAVE_NVTOOLSEXT
+#include "nvToolsExt.h"
+#endif
+
+// Std library headers
+#if defined(CARE_DEBUG) && !defined(_WIN32)
+#include <execinfo.h>
+#endif // defined(CARE_DEBUG) && !defined(_WIN32)
+
+#include <unordered_set>
```
adayton1, line 21: I think this include and several others may no longer be needed.

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,46 @@
...
+
+	class ProfilePlugin : public RAJA::util::PluginStrategy
+	{
+		public:
+			ProfilePlugin();
+
+			char * fileName;
```
adayton1, line 25: Where are the fileName and lineNumber set?

.radiuss-ci
adayton1, line 1: You committed changes here again. Before you do a commit you always need to do a git status to be exactly aware of what you are committing.

tpl/cub
adayton1, line 1: You need to revert this change. I'm guessing you forgot to do "git submodule update --init" when you cloned the CARE repo.

src/care/ProfilePlugin.h
adayton1: This is still not being set anywhere.

src/care/RAJAPlugin.h
```diff
@@ -28,6 +28,7 @@ namespace chai {
 namespace care {
    class RAJAPlugin {
```
adayton1, line 29: We should be able to remove this class now that the functionality has been moved elsewhere.

src/care/ProfilePlugin.h
neelakausik: i think those variables have since been deleted

adayton1: We still need the file and line number. Take a look at what you commented out in forall.h and think about how you can get the file and line number to this class.

src/care/forall.h
```diff
@@ -75,9 +76,11 @@ namespace care {
       const int length = end - start;
 
       if (length != 0) {
-#ifndef CARE_DISABLE_RAJAPLUGIN
-         RAJAPlugin::pre_forall_hook(ExecutionPolicyToSpace<ExecutionPolicy>::value, fileName, lineNumber);
-#endif
+         ProfilePlugin::setFileName(fileName);
```
adayton1, line 79: I would make another class with shared plugin data that both the debug and profile plugin need, such as the file name and line number.

src/care/forall.h
```diff
@@ -491,15 +488,18 @@ namespace care {
    ////////////////////////////////////////////////////////////////////////////////
    template <typename LB, typename Exec>
    void launch_2D_jagged(Exec /*policy*/, int xstart, int /*xend*/, int const * host_lengths, int ystart, int ylength, const char * fileName, int lineNumber, LB && body) {
-      care::RAJAPlugin::pre_forall_hook(chai::CPU, fileName, lineNumber);
+      ProfilePlugin::setFileName(fileName);
```
adayton1, line 491: Did you notice that RAJA::forall is not actually called in this method? That means the RAJA plugins won't actually be called. We need to either manually set the chai execution space or figure out how to use RAJA for 2D kernels.

neelakausik: changed it to set execution space manually for now

src/care/CHAICallback.cpp
```diff
@@ -8,7 +8,9 @@
 // CARE headers
 #include "care/CHAICallback.h"
 #ifndef CARE_DISABLE_RAJAPLUGIN
-#include "care/RAJAPlugin.h"
+#include "care/PluginData.h"
+#include "care/ProfilePlugin.h"
```
adayton1, line 12: ProfilePlugin.h and DebugPlugin.h don't need to be included.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,66 @@
+#include "care/ProfilePlugin.h"
+
+#include "chai/ArrayManager.hpp"
```
adayton1, line 3: I think you can remove the include statements for ArrayManager.hpp, ExecutionSpaces.hpp, and CHAICallback.h.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,66 @@
...
+   unsigned int ProfilePlugin::s_current_color = 0;
+   uint32_t ProfilePlugin::s_colors[7] = { 0x0000ff00, 0x000000ff, 0x00ffff00, 0x00ff00ff, 0x0000ffff, 0x00ff0000, 0x00ffffff };
+   int ProfilePlugin::s_num_colors = sizeof(s_colors) / sizeof(uint32_t);
+   bool ProfilePlugin::s_profile_host_loops = true; 
+   
+   ProfilePlugin::ProfilePlugin() {}
+
+   void ProfilePlugin::preLaunch(const RAJA::util::PluginContext& p) {
+#if CARE_HAVE_NVTOOLSEXT
+#if defined(CARE_GPUCC)
```
adayton1, line 26: This was not your fault - the preexisting code had a bug - but would you be willing to change `CARE_GPUCC` to `__CUDACC__` in ProfilePlugin::preLaunch and ProfilePlugin::postLaunch? `CARE_GPUCC` implies ROCm/Hip support, but NVTX is only supported by CUDA.

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,31 @@
+#ifndef _CARE_ProfilePlugin_H_
+#define _CARE_ProfilePlugin_H_
+
+#include "RAJA/util/PluginStrategy.hpp"
+
+namespace care{
+
+	class ProfilePlugin : public RAJA::util::PluginStrategy
```
adayton1, line 8: Could you try the instructions in this article to fix the formatting of DebugPlugin.h and ProfilePlugin.h using vim? https://vim.fandom.com/wiki/Fix_indentation

A prerequisite is that you modified the .vimrc file to use 3 spaces per tab and convert tabs to spaces. I believe I gave you the instructions on how to do that, but let me know if you need them again.

src/care/forall.h
```diff
@@ -491,15 +485,16 @@ namespace care {
...
+      arrayManager->setExecutionSpace(ExecutionPolicyToSpace<Exec>::value);
+
       // intentional trigger of copy constructor for CHAI correctness
       LB body_to_call{body};
       for (int y = ystart; y < ylength; ++y) {
          for (int x = xstart ; x < host_lengths[y]; ++x) {
             body_to_call(x, y);
          }
       }
-      care::RAJAPlugin::post_forall_hook(chai::CPU, fileName, lineNumber);
```
adayton1, line 502: You'll need a call to `arrayManager->setExecutionSpace(NONE);` here.

src/care/forall.h
```diff
@@ -545,13 +540,14 @@ namespace care {
...
-          care::RAJAPlugin::pre_forall_hook(chai::GPU, fileName, lineNumber);
+          chai::ArrayManager* arrayManager = chai::ArrayManager::getInstance();
+          arrayManager->setExecutionSpace(chai::GPU);
+
           dim3 dimBlock(CARE_CUDA_BLOCK_SIZE, 1);
           dim3 dimGrid;
           dimGrid.x  = (xend/CARE_CUDA_BLOCK_SIZE)+(xend%CARE_CUDA_BLOCK_SIZE==0?0:1);
           dimGrid.y = ylength;
           care_kernel_2D<<<dimGrid, dimBlock>>>( body, gpu_lengths, ylength);
-          care::RAJAPlugin::post_forall_hook(chai::GPU, fileName, lineNumber);
```
adayton1, line 554: You'll need a call to `arrayManager->setExecutionSpace(NONE);` here.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,66 @@
...
+            nvtxRangePop();
+         }
+      }
+#endif
+
+   }
+}
+
+
+static RAJA::util::PluginRegistry::add<care::ProfilePlugin> L ("Care profile plugin", "test");
```
adayton1, line 64: Is "test" just a placeholder?

adayton1, line 64: At this point `ProfilePlugin` is always getting registered with RAJA. I mentioned before that there would be a twist on how to do this. I'd like you to implement a way for users of CARE to control whether `ProfilePlugin` gets registered. It will require adding a static method to the `ProfilePlugin` class.

adayton1: This article might be helpful, particularly the example of using a static variable in a function: https://learn.microsoft.com/en-us/cpp/cpp/storage-classes-cpp?view=msvc-170#static

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace care{
+   unsigned int ProfilePlugin::s_current_color = 0;
+   uint32_t ProfilePlugin::s_colors[7] = { 0x0000ff00, 0x000000ff, 0x00ffff00, 0x00ff00ff, 0x0000ffff, 0x00ff0000, 0x00ffffff };
+   int ProfilePlugin::s_num_colors = sizeof(s_colors) / sizeof(uint32_t);
+   bool ProfilePlugin::s_profile_host_loops = true; 
+   bool ProfilePlugin::registered = false;
+   
+   ProfilePlugin::ProfilePlugin() {}
+   
+   void ProfilePlugin::registerPlugin(bool reg) {
```
adayton1, line 19: The bool argument is misleading since a user might expect they could call this function with true to register the plugin, then later with false to deregister it.

```
void ProfilePlugin::registerPlugin() {
    static RAJA::util::PluginRegistry::add<care::ProfilePlugin> L ("Profile plugin", "Care plugin for profiling"); // This line will only get executed the first time this function is called (that is the nature of a static variable in a function)
}
```

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,59 @@
...
+            nvtxRangePushEx(&eventAttrib);
+         }
+      }
+#endif // CARE_HAVE_NVTOOLSEXT
+#endif // defined(CARE_GPUCC)
+   }
+
+
+   void ProfilePlugin::postLaunch(const RAJA::util::PluginContext& p) {
+#if defined(__CUDACC__)
```
adayton1, line 49: `#if defined(__CUDACC__) && CARE_HAVE_NVTOOLSEXT`

src/care/ProfilePlugin.h
```diff
@@ -0,0 +1,30 @@
...
+#define _CARE_ProfilePlugin_H_
+
+#include "RAJA/util/PluginStrategy.hpp"
+
+namespace care{
+
+   class ProfilePlugin : public RAJA::util::PluginStrategy
+   {
+      public:
+         ProfilePlugin();
```
adayton1, line 11: Best practice if there is nothing in the body of the constructor is to specify `= default`.

e.g. `ProfilePlugin() = default;`

Then you don't have to define an empty constructor in the .cpp file.

src/care/forall.h
```diff
@@ -230,7 +227,6 @@ namespace care {
       forall(RAJA::seq_exec{}, fileName, lineNumber, start, end, std::forward<LB>(body));
 #endif
 
-      RAJAPlugin::setParallelContext(false);
```
adayton1, line 233: `PluginData::setParallelContext(false);`

src/care/forall.h
```diff
@@ -18,11 +18,12 @@
 
 // CARE headers
 #include "care/policies.h"
-#include "care/RAJAPlugin.h"
 #include "care/util.h"
+#include "care/PluginData.h"
 
 // other library headers
 #include "chai/ArrayManager.hpp"
+//#include "chai/ExecutionSpaces.hpp"
```
adayton1, line 26: This should actually be included now.

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
...
+
+         static void removeActivePointer(const chai::PointerRecord* record);
+
+         static void clearActivePointers();
+
+         static int s_threadID;
+
+
+		private:
+         static const char * fileName;
```
adayton1, line 49: Can you follow the style of the other variable names below? The `s_` makes it much easier to see which variables are static.

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace chai {
+   struct PointerRecord;
+}
+
+namespace care{
+
+	class PluginData	{
+		public:
+			PluginData();
```
adayton1, line 17: `PluginData() = default;`

That way you can remove the definition in the .cpp file.

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
+#ifndef _CARE_PluginData_H_
+#define _CARE_PluginData_H_
+
+#include "chai/ExecutionSpaces.hpp"
```
adayton1, line 4: You'll need to clean up your includes (you are missing headers for std::functional, std::unordered_map, and chai::PointerRecord). Also, CHAICallback.h does not need to be included here (it might be needed in PluginData.cpp, though I haven't checked yet).

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,141 @@
...
+            fprintf(callback_output_file, "[CARE] [CHAI] -logchaidata is not supported in this build\n");
+            fflush(callback_output_file);
+#endif // CARE_DEBUG
+         }
+      }
+   }
+}
+
+
+static RAJA::util::PluginRegistry::add<care::DebugPlugin> L ("Debug plugin", "Care plugin for debugging");
```
adayton1, line 140: Can you add a registerPlugin static function to this class, too?

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,141 @@
...
+         GPUWatchpoint::setOrCheckWatchpoint<int>();
+#endif // defined(CARE_GPUCC) && defined(CARE_DEBUG)
+      }
+#endif // !defined(CHAI_DISABLE_RM)
+   }
+
+
+   void DebugPlugin::postLaunch(const RAJA::util::PluginContext& p) {
+#if !defined(CHAI_DISABLE_RM)
+		chai::ExecutionSpace space;
```
adayton1, line 37: Can you fix up the indentation here?

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,141 @@
+#include "care/DebugPlugin.h"
+
+#include "chai/ArrayManager.hpp"
```
adayton1, line 3: I don't think this header is needed.

src/care/RAJAPlugin.cpp
```diff
@@ -1,274 +0,0 @@
...
-            nvtxRangePushEx(&eventAttrib);
-         }
-      }
-#endif // CARE_HAVE_NVTOOLSEXT
-#endif // defined(CARE_GPUCC)
-   }
-
-   /////////////////////////////////////////////////////////////////////////////////
-   ///
-   /// @brief Writes out debugging information after a loop is executed.
```
adayton1, line 108: This doxygen style comment should be moved to the declaration of writeLoopData in DebugPlugin.h

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
...
+
+         static void setParallelContext(bool isParallel);
+
+         static bool isParallelContext();
+
+         static bool post_parallel_forall_action_registered(void * key); 
+
+         static void register_post_parallel_forall_action(void * key, std::function<void(chai::ExecutionSpace, const char *, int)> action);
+
+         static void setSynchronization(bool synchronizeBefore, bool synchronizeAfter);
```
adayton1, line 35: `setSynchronization` doesn't appear to do anything any more. I think it should probably be removed.

adayton1: The variables that it sets can probably be removed, too, and maybe some things that call setSynchronization, too.

adayton1:
I think I've made my last round of comments. Once you've addressed them, then we can do some more testing.

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,143 @@
...
+// Std library headers
+#if defined(CARE_DEBUG) && !defined(_WIN32)
+#include <execinfo.h>
+#endif // defined(CARE_DEBUG) && !defined(_WIN32)
+
+#include <unordered_set>
+
+namespace care{
+   void DebugPlugin::registerPlugin() {
+      static RAJA::util::PluginRegistry::add<care::DebugPlugin> L ("Debug plugin", "Care plugin for debugging");
```
adayton1, line 15: "Care" -> "CARE"

src/care/ProfilePlugin.cpp
adayton1: Did you push up your changes? This doesn't look resolved to me. Also, I prefer that you just make a comment to say if something is fixed and let a reviewer click the "Resolve conversation" button after they have checked that it is resolved.

src/care/PluginData.h
adayton1: This hasn't changed.

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,143 @@
...
+         PluginData::clearActivePointers();
+
+#if defined(CARE_GPUCC) && defined(CARE_DEBUG)
+         GPUWatchpoint::setOrCheckWatchpoint<int>();
+#endif // defined(CARE_GPUCC) && defined(CARE_DEBUG)
+      }
+#endif // !defined(CHAI_DISABLE_RM)
+   }
+
+   /////////////////////////////////////////////////////////////////////////////////
```
adayton1, line 64: Move the doxygen to the header file (it was in the wrong spot originally).

src/care/ProfilePlugin.cpp
neelakausik: got it! also i made all the changes but i think there were problems involving swp files and some of my changes got reverted. i'll look through your comments again to make sure i didn't miss anything else.

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,143 @@
...
+#endif // defined(CARE_DEBUG) && !defined(_WIN32)
+
+            // Flush to the output file
+            fflush(callback_output_file);
+
+#ifdef CARE_DEBUG
+            // Write the arrays captured in the loop
+            usedRecords.clear();
+
+            for (const chai::PointerRecord* record : PluginData::getActivePointers) {
```
adayton1, line 127: getActivePointers should be a function call.

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,56 @@
...
+   uint32_t ProfilePlugin::s_colors[7] = { 0x0000ff00, 0x000000ff, 0x00ffff00, 0x00ff00ff, 0x0000ffff, 0x00ff0000, 0x00ffffff };
+   int ProfilePlugin::s_num_colors = sizeof(s_colors) / sizeof(uint32_t);
+   bool ProfilePlugin::s_profile_host_loops = true; 
+   
+   void ProfilePlugin::registerPlugin() {
+      static RAJA::util::PluginRegistry::add<care::ProfilePlugin> L ("Profile plugin", "CARE plugin for profiling");
+   }
+
+   void ProfilePlugin::preLaunch(const RAJA::util::PluginContext& p) {
+#if CARE_HAVE_NVTOOLSEXT
```
adayton1, line 20: This is checked twice here.

src/care/ProfilePlugin.cpp
adayton1: I think you put the fix in the wrong spot.

src/care/RAJAPlugin.cpp
```diff
@@ -1,274 +0,0 @@
...
-#endif // defined(CARE_GPUCC) && defined(CARE_DEBUG)
-      }
-
-      if (s_update_chai_execution_space) {
-         chai::ArrayManager* arrayManager = chai::ArrayManager::getInstance();
-         arrayManager->setExecutionSpace(chai::NONE);
-      }
-#endif // !defined(CHAI_DISABLE_RM)
-
-      if (s_parallel_context) {
```
adayton1, line 223: This section is missing from DebugPlugin::postLaunch.

src/care/PluginData.h
```diff
@@ -0,0 +1,61 @@
+#ifndef _CARE_PluginData_H_
+#define _CARE_PluginData_H_
+
+#include "chai/PointerRecord.hpp"
+#include "care/config.h"
+#include "care/CHAICallback.h"
```
adayton1, line 6: CHAICallback.h is not needed here

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,55 @@
...
+            eventAttrib.size = NVTX_EVENT_ATTRIB_STRUCT_SIZE;
+            eventAttrib.colorType = NVTX_COLOR_ARGB;
+            eventAttrib.color = s_colors[color_id];
+            eventAttrib.messageType = NVTX_MESSAGE_TYPE_ASCII;
+            eventAttrib.message.ascii = name.c_str();
+            nvtxRangePushEx(&eventAttrib);
+         }
+      }
+#endif // CARE_HAVE_NVTOOLSEXT
+#endif // defined(CARE_GPUCC)
```
adayton1, line 40: I think you have an extra `#endif` here.

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,134 @@
...
+#endif
+         default:
+            space = chai::NONE;
+      }
+
+      if (CHAICallback::isActive()) {			
+         writeLoopData(space, PluginData::getFileName(), PluginData::getLineNumber());
+	
+         // Clear out the captured arrays
+         PluginData::clearActivePointers();
```
adayton1, line 55: I just realized that we only need to clear out active pointers in preLaunch. It doesn't have to happen again in postLaunch.

src/care/DebugPlugin.cpp
adayton1: Thanks!

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,134 @@
...
+   }
+
+   void DebugPlugin::preLaunch(const RAJA::util::PluginContext& p) {
+#if !defined(CHAI_DISABLE_RM)
+      // Prepare to record CHAI data
+      if (CHAICallback::isActive()) {
+         PluginData::clearActivePointers();
+
+#if defined(CARE_GPUCC) && defined(CARE_DEBUG)
+         GPUWatchpoint::setOrCheckWatchpoint<int>();
```
adayton1, line 25: @robinson96, can we remove the GPUWatchpoint stuff? Since we are not passing any address to it, I don't see how it can do anything useful.

adayton1: Oh, I see - you'd have to manually set a watchpoint before the loop. But I looked it up and cuda-gdb at least has the ability to set a watchpoint on both host and GPU memory. So I don't think we need this watchpoint code.

adayton1: Oh, wait. cuda-gdb only supports watchpoints on host memory. Sad day.

adayton1: The question still remains, though, is this used enough to keep it around? I've never used it myself.

adayton1: Looks like rocgdb supports watchpoints on device memory: https://docs.amd.com/projects/ROCdbgapi/en/docs-5.0.0/doxygen/docBin/html/group__watchpoint__group.html

src/care/DebugPlugin.cpp
```diff
@@ -0,0 +1,139 @@
...
+   }
+
+
+   void DebugPlugin::postLaunch(const RAJA::util::PluginContext& p) {
+#if !defined(CHAI_DISABLE_RM)
+	   chai::ExecutionSpace space;
+
+	   switch (p.platform) {
+         case RAJA::Platform::host:
+         	space = chai::CPU; break;
```
liu15, line 38: The indentation is strange in this section.  Are there tabs or is this the case of a beautifier not understanding #ifdefs?

src/care/PluginData.h
```diff
@@ -0,0 +1,64 @@
...
+   struct PointerRecord;
+}
+
+namespace care{
+
+	class PluginData	{
+		public:
+			PluginData() = default;
+
+         static void setFileName(const char * name);
```
dtaller, line 20: I think that the indentation is inconsistent here, between PluginData() = default; and the functions below

src/care/PluginData.h
```diff
@@ -0,0 +1,64 @@
...
+#include <functional>
+#include <unordered_map>
+
+namespace chai {
+   struct PointerRecord;
+}
+
+namespace care{
+
+	class PluginData	{
```
dtaller, line 16: Maybe add a comment describing the purpose of this class

src/care/PluginData.h
```diff
@@ -0,0 +1,64 @@
...
+         
+         static int getLineNumber();   
+
+         static void setParallelContext(bool isParallel);
+
+         static bool isParallelContext();
+
+         static bool post_parallel_forall_action_registered(void * key); 
+
+         static std::unordered_map<void *, std::function<void(chai::ExecutionSpace, const char *, int)>> get_post_parallel_forall_actions();
```
dtaller, line 34: Add a comment here, since a map with a key of a void* is a bit unusual. You might also consider having something like "using ActionMap = std::unordered_map<void *, std::function<void(chai::ExecutionSpace, const char *, int)>>" or whatever since you use it in a few places

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "care/PluginData.h"
+
+/* CUDA profiling macros */
+#if defined(__CUDACC__) && CARE_HAVE_NVTOOLSEXT
+#include "nvToolsExt.h"
+#endif
+
+namespace care{
+   unsigned int ProfilePlugin::s_current_color = 0;
+   uint32_t ProfilePlugin::s_colors[7] = { 0x0000ff00, 0x000000ff, 0x00ffff00, 0x00ff00ff, 0x0000ffff, 0x00ff0000, 0x00ffffff };
```
dtaller, line 11: Comments since many people don't know whether 0x000000ff is purple or orange or whatever

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,54 @@
...
+   int ProfilePlugin::s_num_colors = sizeof(s_colors) / sizeof(uint32_t);
+   bool ProfilePlugin::s_profile_host_loops = true; 
+   
+   void ProfilePlugin::registerPlugin() {
+      static RAJA::util::PluginRegistry::add<care::ProfilePlugin> L ("Profile plugin", "CARE plugin for profiling");
+   }
+
+   void ProfilePlugin::preLaunch(const RAJA::util::PluginContext& p) {
+#if defined(__CUDACC__) && CARE_HAVE_NVTOOLSEXT
```
dtaller, line 20: Will something different be needed for HIP ? Maybe add a TODO ?

src/care/PluginData.h
```diff
@@ -0,0 +1,64 @@
...
+         static void addActivePointer(const chai::PointerRecord* record);
+
+         static void removeActivePointer(const chai::PointerRecord* record);
+
+         static void clearActivePointers();
+
+         static int s_threadID;
+
+
+		private:
```
dtaller, line 51: More indentation issues

dtaller:
Could you add a pull request description at the top?

src/care/ProfilePlugin.cpp
adayton1: Or just add a TODO comment saying that we need label the colors.

src/care/ProfilePlugin.cpp
adayton1: The Hip interface is similar, but we may want to switch to using caliper or even let the host code provide callbacks for profiling. So for now I would just add a comment that profiling is only supported in CUDA builds.

dtaller: That sounds fine. Would you say just a comment is appropriate, or would you suggest a warning or log message to the user, or even a failure/exit?

src/care/PluginData.h
adayton1: I think this was added by Peter, so we should have him add the comment. I'm not even sure what this is used for, but I think it's something to do with race condition checking.

The alias is a good idea, though.

dtaller:
Also, care builds some internal tests (see care/test/ folder ). Do you think any tests are needed? Or would it be difficult to add unit tests for this type of feature?

src/care/ProfilePlugin.cpp
adayton1: We could output a warning in the registerPlugin function if `__HIPCC__` is defined. I don't think we want a failure/exit.

adayton1: Here's an example of a CARE warning message: https://github.com/LLNL/CARE/blob/8a08fb49ce3ef93837a14595daf4f18a0584eecf/src/care/scan_impl.h#L49

src/care/ProfilePlugin.cpp
```diff
@@ -0,0 +1,63 @@
...
+                                           0x00ff00ff,   //pink
+                                           0x0000ffff,   //cyan
+                                           0x00ff0000,   //red
+                                           0x00ffffff }; //white
+   int ProfilePlugin::s_num_colors = sizeof(s_colors) / sizeof(uint32_t);
+   bool ProfilePlugin::s_profile_host_loops = true; 
+   
+   void ProfilePlugin::registerPlugin() {
+#if !defined(__HIPCC_)
+      printf("[CARE] Warning: Profiling is only supported in CUDA builds.\n");
```
dtaller, line 23: I think you made a typo here in the #if define, unless I am missing something. This should probably be #if defined(__HIPCC__) ? It gives this warning if hip is defined (you have it if hip is not defined I think, with the !, and you deleted one of the underscored after HIPCC)?

neelakausik: yes you're right! i will fix that

src/care/DebugPlugin.cpp
adayton1: I guess we can leave the support in for now and it would be easy to remove later if we decide to get rid of it.

src/care/PluginData.h
```diff
@@ -0,0 +1,63 @@
...
+#include <vector>
+#include <functional>
+#include <unordered_map>
+
+namespace chai {
+   struct PointerRecord;
+}
+
+namespace care{
+   //class for plugin functions and variables
```
adayton1, line 15: Class for shared plugin functions and variables

adayton1:
> Could you add a pull request description at the top?

I expanded on Neela's description. Let me know if it's not clear.

src/care/DebugPlugin.cpp
adayton1: I think this still needs to be fixed. And I'll give myself a TODO to figure out how to add a beautifier to standalone CARE.

liu15 (approved):

dtaller:
Thanks for your work!

I think that all my comments were addressed except for the alias for the strange map "using ActionMap = std::unordered_map<void *, std::function<void(chai::ExecutionSpace, const char *, int)>>". Either make the change or tell me if you ultimately decided not to do that, and then I'll approve

src/care/PluginData.h
neelakausik: just added the alias!

dtaller (approved):

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
...
+         static int s_threadID;
+
+      private:
+         static const char * s_file_name;
+
+         static int s_line_number;
+
+         static bool s_parallel_context;
+
+         static std::unordered_map<void *, std::function<void(chai::ExecutionSpace, const char *, int)>> s_post_parallel_forall_actions;
```
dtaller, line 59: Sorry one more comment :D . Is this also an ActionMap, or is the signature a bit different?

neelakausik: yes you're right, i missed that!

dtaller: No worries! Thanks for making the fix!

src/care/PluginData.h
```diff
@@ -0,0 +1,65 @@
...
+#include "care/config.h"
+#include <vector>
+#include <functional>
+#include <unordered_map>
+
+namespace chai {
+   struct PointerRecord;
+}
+
+using ActionMap = std::unordered_map<void *, std::function<void(chai::ExecutionSpace, const char *, int)>>;
```
adayton1, line 14: Move the alias inside the care namespace.

src/care/DebugPlugin.cpp
adayton1: @neelakausik, can you fix the indentation?

adayton1: Thanks!

adayton1 (approved):
LGTM

PR #227, opened 2023-07-12
Revert to BLT v0.5.2

adayton1:
BLT v0.5.3 has some issues when building gtest on rzansel. Reverting to v0.5.2.


neelakausik (approved):

scripts/spack_packages/packages/care/package.py
```diff
@@ -40,7 +40,7 @@ class Care(CachedCMakePackage, CudaPackage, ROCmPackage):
     depends_on('cmake@3.14.5:', type='build')
     depends_on('cmake@3.21:', when='+rocm', type='build')
 
-    depends_on('blt@0.5.2:', type='build', when='@0.10.0:')
+    depends_on('blt@0.5.2', type='build', when='@0.10.0:')
```
liu15, line 43: I'm not familiar with the spack syntax.  What does the colon mean?

liu15 (approved):

scripts/spack_packages/packages/care/package.py
dtaller: I think `blt@0.5.2:` would mean that spack can use blt 0.5.2 or greater (at least that version), while without the colong we we fixed to version 0.5.2 exactly

dtaller (approved):
Approved. If you have not already, please let the BLT team know about the issues.

PR #229, opened 2023-07-19
Make BenchmarkForall more repeatable

adayton1:
* Add warmup kernels
* Add synchronizes where needed

### DIFF
--- a/benchmarks/BenchmarkForall.cpp
+++ b/benchmarks/BenchmarkForall.cpp
@@ -20,6 +20,7 @@ static void benchmark_sequential_loop(benchmark::State& state) {
    const int size = state.range(0);
    care::host_device_ptr<int> data(size, "data");
 
+   // For consistency with the GPU case, which requires a warm up kernel
    CARE_SEQUENTIAL_LOOP(i, 0, size) {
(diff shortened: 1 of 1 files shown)


neelakausik (approved):

benchmarks/BenchmarkForall.cpp
```diff
@@ -37,10 +42,18 @@ static void benchmark_openmp_loop(benchmark::State& state) {
    const int size = state.range(0);
    care::host_device_ptr<int> data(size, "data");
 
+   CARE_OPENMP_LOOP(i, 0, size) {
```
dtaller, line 45: I would add a comment on why this is being done here (I think consistency with the GPU test, where you are doing it to "warm up kernel", to avoid startup/initialization issues that made timing inconsistent).

dtaller (approved):
Approving, just address my suggestion to add a comment before merging though

liu15 (approved):

benchmarks/BenchmarkForall.cpp
adayton1: Fixed.

PR #230, opened 2023-07-19
Add a resource with custom allocator

adayton1:
* This is an experimental resource that could make RAJA allocations thread safe and stream aware

### DIFF
--- a/test/TestCudaUmpireResource.cpp
+++ b/test/TestCudaUmpireResource.cpp
@@ -11,22 +11,60 @@
 
 // other library headers
 #include "gtest/gtest.h"
+#include "umpire/strategy/QuickPool.hpp"
 
 // care headers
 #include "care/CudaUmpireResource.h"
-#include "care/DefaultMacros.h"
-#include "care/policies.h"
 #include "care/detail/test_utils.h"
 
 GPU_TEST(CudaUmpireResource, gpu_initialization) {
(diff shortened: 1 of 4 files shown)


test/TestCudaUmpireResource.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+GPU_TEST(CudaUmpireResource, DefaultConstructor)
+{
+   care::CudaUmpireResource resource;
+}
+
+GPU_TEST(CudaUmpireResource, AllocatorConstructor)
+{
+   auto& rm = umpire::ResourceManager::getInstance();
```
adayton1, line 31: @neelakausik, essentially you will need to create a device pool, a pinned pool, and a unified memory pool per thread, then use those to create a resource per thread. The call to init_care_for_testing() above creates a GPU wide pool called "DEVICE_POOL" and another pool called "PINNED_POOL". Those should be the basis for the per thread device and pinned memory pools. On line 53 I create a GPU wide unified memory pool called "UM_POOL". You will need to create that manually in whatever test cases you have, then have that allocator be the basis for the per thread unified memory pool.

adayton1: You should test out this functionality in standalone CARE. Write some tests and benchmarks that involve RAJA reducers, scans, and CARE loops.

PR #232, opened 2023-07-19
added forall_with_streams and updated BenchmarkForall.cpp

neelakausik:


adayton1:
Didn't you have a macro in DefaultMacros.h as well?

benchmarks/BenchmarkForall.cpp
```diff
@@ -69,6 +69,23 @@ static void benchmark_gpu_loop(benchmark::State& state) {
 // Register the function as a benchmark
 BENCHMARK(benchmark_gpu_loop)->Range(1, INT_MAX);
 
+static void benchmark_gpu_loop_streams(benchmark::State& state) {
```
adayton1, line 72: I'd like to see more benchmarks - especially one that uses threads to launch kernels in different streams.

benchmarks/BenchmarkForall.cpp
```diff
@@ -69,6 +69,23 @@ static void benchmark_gpu_loop(benchmark::State& state) {
...
+   const int size = state.range(0);
+   care::host_device_ptr<int> data(size, "data");
+
+   for (auto _ : state) {
+      RAJA::resources::Cuda res;
+      care::forall_with_stream(care::gpu{}, res, "BenchmarkForall.cpp", 78, 0, size, [=] CARE_DEVICE (int i) {
+         data[i] = i;
```
adayton1, line 79: This is a data race.

benchmarks/BenchmarkForall.cpp
adayton1: Where are the ones you used for your poster?

neelakausik: these are the ones i used, but i can use the ones i made earlier with multiple kernels

src/care/DefaultMacros.h
```diff
@@ -548,6 +548,15 @@
 #define CARE_CHECKED_PARALLEL_LOOP_END(CHECK) }); \
    CARE_NEST_END(CHECK) }}
 
+#define CARE_CHECKED_STREAMED_LOOP_START(INDEX, RESOURCE, START_INDEX, END_INDEX, CHECK) { \
```
adayton1, line 551: Can you make RESOURCE the first argument of this macro (and the CARE_STREAMED_LOOP macro below)?

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+#define NUM_KERNELS 4
+
+using namespace care;
+
+#if defined(CARE_GPUCC)
+//each kernel has a separate stream
+static void benchmark_gpu_loop_separate_streams(benchmark::State& state) {
+   int N = state.range(0);
+	const char * fileName = "test";
```
adayton1, line 31: Can you fix the indentation in this file?

src/care/forall.h
```diff
@@ -189,6 +189,49 @@ namespace care {
...
+               const int start, const int end, LB&& body) {
+#if CARE_ENABLE_PARALLEL_LOOP_BACKWARDS
+      s_reverseLoopOrder = true;
+#endif
+
+#if CARE_ENABLE_GPU_SIMULATION_MODE
+      forall(gpu_simulation{}, res, fileName, lineNumber, start, end, std::forward<LB>(body));
+#elif defined(__CUDACC__)
+      forall(RAJA::cuda_exec<CARE_CUDA_BLOCK_SIZE, CARE_CUDA_ASYNC>{},
+             res, RAJA::RangeSegment(start, end), std::forward<LB>(body));
```
adayton1, line 221: I think you need an overload of forall that takes a resource, right?

src/care/DefaultMacros.h
```diff
@@ -548,6 +548,15 @@
 #define CARE_CHECKED_PARALLEL_LOOP_END(CHECK) }); \
    CARE_NEST_END(CHECK) }}
 
+#define CARE_CHECKED_STREAMED_LOOP_START(RESOURCE, INDEX, START_INDEX, END_INDEX, CHECK) { \
```
adayton1, line 551: If you search for CARE_CHECKED_PARALLEL_LOOP_START, you'll see that it is defined one of two ways, depending on whether we have passed an option to use a legacy mode. You'll need to add another definition for CARE_CHECKED_STREAMED_LOOP_START/END in the legacy mode case.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,85 @@
...
+   RAJA::resources::Event event_arr[NUM_KERNELS];
+   for(int i = 0; i < NUM_KERNELS; i++)
+   {
+      RAJA::resources::Cuda res;
+      res_arr[i] = res;
+      RAJA::resources::Event e = res.get_event();
+      event_arr[i] = e;
+   }
+	
+   care::host_device_ptr<int> arr(N, "arr");
```
adayton1, line 42: You need one array for each kernel.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <cmath>
+
+#define NUM_KERNELS 4
+
+using namespace care;
+
+#if defined(CARE_GPUCC)
+//each kernel has a separate stream
+static void benchmark_gpu_loop_separate_streams(benchmark::State& state) {
+   int N = state.range(0);
```
adayton1, line 30: Instead of making N the size of the array, I would pick a fixed size for the array (maybe 100,000 or 1,000,000), then use N as the number of kernels.

`BENCHMARK(benchmark_gpu_loop_separate_streams)->Arg(1)->Arg(2)->Arg(4)->Arg(8)->Arg(12)->Arg(16);`

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,85 @@
...
+   for(int i = 0; i < NUM_KERNELS; i++)
+   {
+      RAJA::resources::Cuda res;
+      res_arr[i] = res;
+      RAJA::resources::Event e = res.get_event();
+      event_arr[i] = e;
+   }
+	
+   care::host_device_ptr<int> arr(N, "arr");
+   for (auto _ : state) {
```
adayton1, line 43: It's helpful to have a warmup kernel before going into this loop.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,85 @@
...
+      //run num kernels
+      for(int j = 0; j < NUM_KERNELS; j++)
+      {
+         CARE_STREAMED_LOOP(res_arr[j], i, 0 , N) {
+         arr[i] = i;
+         } CARE_STREAMED_LOOP_END					
+         if(j > 0) res_arr[j].wait_for(&event_arr[j - 1]);
```
adayton1, line 50: Why are you waiting for the previous kernel to finish? Won't that make them happen one after the other rather than potentially at the same time?

src/care/forall.h
neelakausik: what do you mean?

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#define size 1000000
+
+namespace care{
+
+#if defined(CARE_GPUCC)
+//each kernel has a separate stream
+static void benchmark_gpu_loop_separate_streams(benchmark::State& state) {
+   int N = state.range(0);
+   RAJA::resources::Cuda res_arr[N];
```
adayton1, line 31: Declaring a fixed size array like this with a run time variable is not supported by all compilers. You could hard code it to 16 since that is the max size.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+   RAJA::resources::Event event_arr[N];
+   care::host_device_ptr<int> arrays[16];
+   for(int i = 0; i < N; i++)
+   {
+      RAJA::resources::Cuda res;
+      res_arr[i] = res;
```
adayton1, line 37: You can do the creation and assignment all on one line:
`res_arr[i] = RAJA::resources::Cuda();`

Same thing for the other arrays.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+      RAJA::resources::Event e = res.get_event();
+      event_arr[i] = e;
+      care::host_device_ptr<int> arr(size, "arr");
+      arrays[i] = arr;
+   }
+
+   //warmup kernel
+   RAJA::resources::Cuda warmup_res;
+   CARE_STREAMED_LOOP(warmup_res, i, 0 , size) {
```
adayton1, line 46: Just use a regular loop without a resource for the warm up kernel. You'll also need to do a device synchronize after the warm up kernel to make sure it doesn't affect the timings.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+   for (auto _ : state) {
+      //run num kernels
+      for(int j = 0; j < N; j++)
+      {
+         CARE_STREAMED_LOOP(res_arr[j], i, 0 , size) {
+            arrays[j][i] = i;
+         } CARE_STREAMED_LOOP_END					
+      }
+   }
+   for(int i = 0; i < N; i++) {arrays[i].free();}
```
adayton1, line 59: Just a stylistic note - I prefer that the body of a for loop is on a separate line.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+#if defined(CARE_GPUCC)
+//each kernel has a separate stream
+static void benchmark_gpu_loop_separate_streams(benchmark::State& state) {
+   int N = state.range(0);
+   RAJA::resources::Cuda res_arr[N];
+   RAJA::resources::Event event_arr[N];
+   care::host_device_ptr<int> arrays[16];
+   for(int i = 0; i < N; i++)
+   {
+      RAJA::resources::Cuda res;
```
adayton1, line 36: You'll need to use a type alias that could either be Cuda or Hip.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+   } CARE_STREAMED_LOOP_END					
+	
+   for (auto _ : state) {
+      //run num kernels
+      for(int j = 0; j < N; j++)
+      {
+         CARE_STREAMED_LOOP(res_arr[j], i, 0 , size) {
+            arrays[j][i] = i;
+         } CARE_STREAMED_LOOP_END					
+      }
```
adayton1, line 57: Is there a wait all function that could wait for the events on all the resources to complete? Your previous attempt made all the loops execute one after the other. Now they are launching in parallel, but the timing does not include the actual execution.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <benchmark/benchmark.h>
+#include <omp.h>
+
+// Std library headers
+#include <climits>
+#include <cmath>
+
+#define size 1000000
+
+namespace care{
```
adayton1, line 25: Benchmarks are only cpp files, so they don't need to be inside the CARE namespace (if there were header files, then yes, they would need to be in a namespace).

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+   //warmup kernel
+   RAJA::resources::Cuda warmup_res;
+   CARE_STREAMED_LOOP(warmup_res, i, 0 , size) {
+      arrays[0][i] = 0;
+   } CARE_STREAMED_LOOP_END					
+	
+   for (auto _ : state) {
+      //run num kernels
+      for(int j = 0; j < N; j++)
```
adayton1, line 52: Shouldn't this be an OpenMP loop?

benchmarks/BenchmarkRaja.cpp
adayton1: I think there is a care::gpuDeviceSynchronize method that will work for both HIP and CUDA.

benchmarks/BenchmarkRaja.cpp
adayton1: If there's not a wait all function, then you'll need to do a device synchronize.

src/care/forall.h
neelakausik: I thought it was calling RAJA::forall, but will look into it further

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,108 @@
...
+   } CARE_GPU_LOOP_END
+
+   care::gpuDeviceSynchronize(__FILE__, __LINE__);
+
+   for (auto _ : state) {
+      //run num kernels
+      #pragma omp parallel for
+      for(int j = 0; j < N; j++)
+      {
+         CARE_STREAMED_LOOP(res, i, 0, size) {
```
adayton1, line 90: Don't use a resource here - we want everything in this case to happen on the default stream.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,108 @@
...
+            arrays[j][i] = i;
+         } CARE_STREAMED_LOOP_END
+      }
+      res.wait();
```
adayton1, line 94: Do a device synchronize here instead of wait - that will make it a more fair comparison to the other benchmark with streams.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,108 @@
...
+   //warmup kernel
+   CARE_GPU_LOOP(i, 0, size) {
+      arrays[0][i] = i;
+   } CARE_GPU_LOOP_END
+
+   care::gpuDeviceSynchronize(__FILE__, __LINE__);
+
+   for (auto _ : state) {
+      //run num kernels
+      #pragma omp parallel for
```
adayton1, line 87: I don't think we really want to use OpenMP here - we want this benchmark to match as closely as possible our current way of doing things, and we don't use OpenMP to launch a bunch of kernels on a single stream.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+   care::gpuDeviceSynchronize(__FILE__, __LINE__);   
+	
+   for (auto _ : state) {
+      //run num kernels
+      #pragma omp parallel for
+      for(int j = 0; j < N; j++)
+      {
+         CARE_STREAMED_LOOP(res_arr[j], i, 0 , size) {
+            arrays[j][i] = i;
```
adayton1, line 52: It would probably be worth actually doing some work in this kernel (square roots, trig functions, etc...). If the kernel basically does nothing, then you won't see a win from overlapping.

src/care/forall.h
adayton1: It shouldn't be calling RAJA::forall - there's no overload that takes the fileName and lineNumber.

adayton1: Also, that's the main reason I dislike "using namespace..." statements - it's too easy to accidentally call the wrong function.

benchmarks/BenchmarkRaja.cpp
```diff
@@ -0,0 +1,106 @@
...
+   //warmup kernel
+   CARE_GPU_LOOP(i, 0 , size) {
+      arrays[0][i] = 0;
+   } CARE_GPU_LOOP_END	
+
+   care::gpuDeviceSynchronize(__FILE__, __LINE__);   
+	
+   for (auto _ : state) {
+      //run num kernels
+      omp_set_num_threads(16);
```
adayton1, line 48: This should be N rather than 16.

src/care/forall.h
```diff
@@ -97,6 +101,50 @@ namespace care {
...
+   /// @arg[in] res Resource to be used
+   /// @arg[in] fileName The name of the file where this function is called
+   /// @arg[in] lineNumber The line number in the file where this function is called
+   /// @arg[in] start The starting index (inclusive)
+   /// @arg[in] end The ending index (exclusive)
+   /// @arg[in] body The loop body to execute at each index
+   ///
+   ////////////////////////////////////////////////////////////////////////////////
+   template <typename ExecutionPolicy, typename LB>
+   void forall(ExecutionPolicy /* policy */, Resource res, const char * fileName, const int lineNumber,
```
adayton1, line 121: You should add a template parameter as the resource type since we want to be able to accept any resource type.

PR #243, opened 2024-01-25
Enable syncIfNeeded when resource management is disabled (needed for …

liu15:
…single-memory-space mode)


adayton1:
Included with #246

PR #248, opened 2024-03-09
Support for single memory space and other fixes

adayton1:
* Add support for single memory space
* Windows shared library build fixes
* Support for unsigned and 64 bit integers in algorithms and scans
* Optimizations to reduce host side memory touches


adayton1:
I'm leaving out the gpu sim mode fixes until after the release (I think there are some updates to CHAI needed as well).

liu15 (approved):
Thanks for upstreaming all of this!

dtaller (approved):

PR #261, opened 2024-05-23
Add config file and example for thread sanitizer support

adayton1:
Demonstrates that thread sanitizer works with RAJA/CARE


dtaller (approved):

examples/RaceConditions.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "care/algorithm.h"
+
+// Std library headers
+#include <algorithm>
+#include <iostream>
+#include <limits>
+#include <vector>
+
+int main(int, char**) {
+   int size = 10;
```
dtaller, line 19: this is nitpicky, but you could have size as a constant

adayton1: Fixed.

liu15 (approved):

PR #278, opened 2024-07-31
Fix chunked loop macros

adayton1:
* Make sure top-level chunked loop macros call the appropriate underlying macros
* Fix chunked loop test in TestForall

### DIFF
--- a/src/care/DefaultMacros.h
+++ b/src/care/DefaultMacros.h
@@ -313,7 +313,7 @@ OMP_FOR_BEGIN for (auto INDEX = _care_openmp_for_loop_chunk_begin_ndx; INDEX < _
 /// @arg[in] CHECK The variable to check that the start and end macros match
 ///
 ////////////////////////////////////////////////////////////////////////////////
-#define CARE_CHECKED_CHUNKED_GPU_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK) CARE_CHECKED_OPENMP_FOR_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK)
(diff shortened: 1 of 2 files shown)


src/care/DefaultMacros.h
```diff
@@ -313,7 +313,7 @@ OMP_FOR_BEGIN for (auto INDEX = _care_openmp_for_loop_chunk_begin_ndx; INDEX < _
 /// @arg[in] CHECK The variable to check that the start and end macros match
 ///
 ////////////////////////////////////////////////////////////////////////////////
-#define CARE_CHECKED_CHUNKED_GPU_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK) CARE_CHECKED_OPENMP_FOR_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK)
+#define CARE_CHECKED_CHUNKED_GPU_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK) CARE_CHECKED_CHUNKED_OPENMP_FOR_LOOP_START(INDEX, START_INDEX, END_INDEX, CHUNK_SIZE, CHECK)
```
dtaller, line 316: We are defining GPU loops to be OPENMP loops. This is because we are in a section for CARE_LEGACY_COMPATIBILITY_MODE, correct? Why is this called CARE_LEGACY_COMPATIBILITY_MODE as opposed to CARE_OPENMP or something like that?

dtaller (approved):

liu15 (approved):

src/care/DefaultMacros.h
adayton1: In CARE_LEGACY_COMPATIBILITY_MODE not everything is backed by openmp - the sequential loops are translated to raw for loops, for example.

dtaller: Thanks

PR #285, opened 2024-08-29
Bugfix/probinso/facekey device

robinson96:
Const correctness fix in uniqArray API.

Strictly use '<' in care::BinarySearch.


src/care/algorithm_impl.h
```diff
@@ -471,7 +471,7 @@ CARE_HOST_DEVICE CARE_INLINE int BinarySearch(const T *map, const int start,
 
    while (khi-klo > 1) {
       k = (khi+klo) >> 1 ;
-      if (map[k] == num) {
+      if (! (map[k] < num) && !(num < map[k])) {
```
liu15, line 474: can you add a comment somewhere in this routine indicating that it should only use <?

liu15 (approved):

PR #292, opened 2024-09-13
Fixes for some nvcc versions

dtaller:
Sometimes nvcc won't generate constructors/assignment operators when `= default` is used. The workaround is to explicitly define them.


src/care/host_device_map.h
```diff
@@ -66,7 +66,7 @@ namespace care {
    class host_device_map< key_type, mapped_type, RAJA::seq_exec> {
       public:
          // default constructor
-         host_device_map() noexcept = default;
+         host_device_map() noexcept {};
```
liu15, line 69: What is the issue here?  Are you basically getting rid of the GPU constructor?

dtaller: I got this bug during offsite builds, and this was a fix I discussed with Alan Dayton and Jason Burmarck

/p/work1/taller1/JUNE_2024/nativempi/nocheetah/lib/nautilus_gpu/raja/2024.02.0.a/compiler-gcc-8.5.0/cpp-17/blt-0.6.2.d/cmake-3.21.1/camp-2024.02.0.a/cuda-11.6/mpi-openmpi-5.0.0/cub-1.8.0.c/include/RAJA/util/plugins.hpp(28): error: function "lambda [](size_t)->void::<unnamed>(const lambda [](size_t)->void &)" (declared implicitly) cannot be referenced -- it is a deleted function
          detected during:
            instantiation of "auto RAJA::util::trigger_updates_before(T &&)->std::remove_reference<T>::type [with T=lambda [](size_t)->void &]"

dtaller: Yes, exactly

liu15 (approved):

src/care/host_device_map.h
adayton1: This one should be changed back to `= default`

src/care/host_device_map.h
```diff
@@ -380,7 +386,7 @@ namespace care {
    {
       public:
          // default constructor
-         host_device_map() noexcept = default;         
+         host_device_map() noexcept {};
```
adayton1, line 389: This is host only as well, so these implementations can be changed back to `= default`.

src/care/host_device_map.h
dtaller: Reverted

src/care/host_device_map.h
```diff
@@ -226,7 +226,13 @@ namespace care {
          }
 
          // copy constructor 
-         host_device_map(host_device_map const & other) noexcept = default;
+         CARE_HOST_DEVICE host_device_map(host_device_map const & other) noexcept {
```
adayton1, line 229: Can you use the member initializer format?

```
CARE_HOST_DEVICE host_device_map(host_device_map const & other) noexcept :
   m_max_size(other.m_max_size),
   m_signal(other.m_signal),
   ...
{
}
```

src/care/host_device_map.h
```diff
@@ -399,7 +405,13 @@ namespace care {
          }
          
          // copy constructor 
-         host_device_map(host_device_map const & other) noexcept = default;
+         host_device_map(host_device_map const & other) noexcept {
```
adayton1, line 408: This one is host only, too, so it can be reverted back to `= default;`

dtaller: Reverted

liu15:
Was this the only change needed for the offsite builds?  I thought that =default was supposed to create both host and device default versions.  I'm surprised that these are the only ones...

src/care/host_device_map.h
dtaller: Done

dtaller:
> Was this the only change needed for the offsite builds? I thought that =default was supposed to create both host and device default versions. I'm surprised that these are the only ones...

@liu15 Yes, this only happened on offsite builds. I am not totally sure why, it may even be a compiler problem. But the error occurred during capture of care objects in raja loops.

adayton1 (approved):
LGTM

adayton1:
> > Was this the only change needed for the offsite builds? I thought that =default was supposed to create both host and device default versions. I'm surprised that these are the only ones...
> 
> @liu15 Yes, this only happened on offsite builds. I am not totally sure why, it may even be a compiler problem. But the error occurred during capture of care objects in raja loops.

`= default` is supposed to create both host and device versions of the constructor. But nvcc sometimes fails at generating the device version. The RAJA team has experienced this as well.

PR #303, opened 2024-11-14
Deep copy

liu15:
Add DEEP_COPY_RAW_PTR option for managed array initialization from raw pointers


adayton1 (approved):

adayton1:
Can you add a release note?

dtaller (approved):

dtaller:
This MR was fine, but I do have two questions. Why did you make deep copy a compile time option instead of an argument, as for example

`host_device_ptr<T>(T* from, size_t size, const char * name. bool deepCopy=false)`

Also, what are the use cases for this?

PR #311, opened 2025-01-23
Add device address sanitizer example

adayton1:
* Update BLT_CXX_STD default to c++17
* Update host config to rocm 6.3.0
* Add working device address sanitizer example

### DIFF
--- a/examples/DeviceASAN.cpp
+++ b/examples/DeviceASAN.cpp
@@ -16,6 +16,12 @@ int main(int, char**) {
       data[i] = i;
    } CARE_STREAM_LOOP_END
 
+   // This will write past the end of the array.
+   // ASAN should report something like the following:
+   //
+   // ==4084715==ERROR: AddressSanitizer: heap-buffer-overflow on amdgpu device 0 at pc 0x153d3cdd3db8
+   // READ of size 4 in workgroup id (3906,0,0)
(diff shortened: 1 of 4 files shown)


liu15 (approved):

robinson96 (approved):
LGTM

examples/DeviceASAN.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+int main(int, char**) {
+   const int size = 1000000;
+   care::host_device_ptr<int> data(size);
+
+   CARE_STREAM_LOOP(i, 0, size) {
+      data[i] = i;
+   } CARE_STREAM_LOOP_END
+
+   CARE_STREAM_LOOP(i, 0, size + 1) {
```
dtaller, line 19: I would just add a comment about how this is an intentional bug going one past the end, and also describe what the expected address sanitizer output should look like in response to the intentional bug.

dtaller (approved):
Approving, but an extra comment as I described would be nice

examples/DeviceASAN.cpp
adayton1: Good idea! Done.

dtaller: Thanks!

dtaller: I don't see the comment, don't forget to commit it before you merge.

PR #326, opened 2025-04-02
Miscellaneous bug fixes

adayton1:
* Fix unexpected behavior in care::sort_uniq (a case where the input array should not be modified)
* Fixes care::host_device_map::reserve in the force_keyvaluesorter specialization
* Removes unused variable

### DIFF
--- a/src/care/algorithm_impl.h
+++ b/src/care/algorithm_impl.h
@@ -844,11 +844,16 @@ CARE_INLINE void sortArray(RAJA::seq_exec, care::host_device_ptr<T> &Array, size
 * Function  : sort_uniq(<T>_ptr)
 * Author(s) : Peter Robinson
 * Purpose   : Sorts and uniques an array.
+* @param    : e Execution policy
+* @param    : array: pointer to an array to sort and uniq
+* @param    : len: length of the array
+* @param    : noCopy. If true, implementation is free to store a completely new array at pointer
+*             If false, implementation will not mess with the underlying allocation of the new array
 **************************************************************************/
 template <typename T, typename Exec>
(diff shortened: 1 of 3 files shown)


src/care/algorithm_impl.h
```diff
@@ -844,11 +844,16 @@ CARE_INLINE void sortArray(RAJA::seq_exec, care::host_device_ptr<T> &Array, size
...
+* @param    : len: length of the array
+* @param    : noCopy. If true, implementation is free to store a completely new array at pointer
+*             If false, implementation will not mess with the underlying allocation of the new array
 **************************************************************************/
 template <typename T, typename Exec>
 CARE_INLINE void sort_uniq(Exec e, care::host_device_ptr<T> * array, int * len, bool noCopy)
 {
-   if ((*len) == 0) {
+   if ((*len) == 0 && noCopy) {
       if ((*array) != nullptr) {
```
dtaller, line 857: Maybe I am misunderstanding, but shouldn't the && noCopy be on this line instead? I think if len is 0 and noCopy is true, we still want to return early without calling sortArray or uniqueArray. We just don't want to free the array?

adayton1: That's a good question. @robinson96 made this fix in our fork. Can you weigh in, Peter?

adayton1: My guess is that both sortArray and uniqArray will short circuit if the length is 0, so it should still be correct. But I think your suggestion makes it clearer, Danny.

adayton1: The more I look at this, the more I believe you are right, Danny. So I went ahead and made the change you suggested.

dtaller (approved):

liu15 (approved):

PR #396, opened 2026-09-03
Add segmented sort

adayton1:
A segmented sort is used when many small arrays need to be sorted. They are all put into one contiguous array and offsets to each segment are provided. The algorithm sorts each segment independently.


liu15 (approved):

robinson96 (approved):
LGTM

PR #398, opened 2026-09-04
Fixes for segmented sort

adayton1:
* Fixes test cases (tried and failed to create the same umpire allocator multiple times)
* Made the segments more obvious in the test cases
* Added a test case to exercise slices
* Added documentation to segmented_sort functions

### DIFF
--- a/src/care/cuda/sort.h
+++ b/src/care/cuda/sort.h
@@ -24,7 +24,8 @@ namespace care::cuda {
  * @param keys Keys to sort in place.
  * @param offsets Segment boundaries. For N segments, offsets must contain
  * N + 1 entries: offsets[i] begins segment i, and offsets[N] marks the end of
- * the final segment. Segment i is therefore [offsets[i], offsets[i + 1]).
+ * the final segment. Segment i is therefore [offsets[i], offsets[i + 1]). The
+ * entries must be monotonically nondecreasing and within [0, keys.size()].
  */
 template <typename KeyT, typename OffsetT>
 CARE_INLINE void segmented_sort(care::host_device_ptr<KeyT>& keys,
--- a/src/care/cuda/sort.h
+++ b/src/care/cuda/sort.h
@@ -24,8 +24,10 @@ namespace care::cuda {
  * @param keys Keys to sort in place.
  * @param offsets Segment boundaries. For N segments, offsets must contain
  * N + 1 entries: offsets[i] begins segment i, and offsets[N] marks the end of
- * the final segment. Segment i is therefore [offsets[i], offsets[i + 1]). The
- * entries must be monotonically nondecreasing and within [0, keys.size()].
+ * the final segment. Segment i is therefore [offsets[i], offsets[i + 1]).
+ * The entries must form a nondecreasing sequence from 0 to keys.size(); that
+ * is, offsets[0] must be 0 and offsets[N] must equal keys.size(). Repeated
+ * entries denote empty segments.
  */
 template <typename KeyT, typename OffsetT>
 CARE_INLINE void segmented_sort(care::host_device_ptr<KeyT>& keys,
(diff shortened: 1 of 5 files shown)


dtaller (approved):

src/care/cuda/sort.h
```diff
@@ -19,6 +19,13 @@
 
 namespace care::cuda {
 
+/**
+ * @brief Sort keys in ascending order independently within each segment.
+ * @param keys Keys to sort in place.
+ * @param offsets Segment boundaries. For N segments, offsets must contain
+ * N + 1 entries: offsets[i] begins segment i, and offsets[N] marks the end of
+ * the final segment. Segment i is therefore [offsets[i], offsets[i + 1]).
```
Arlie-Capps: Design question:

Because `segment i is ... [offsets[i], offsets[i + 1])`, the segments are in order, there are no gaps between the segments, and segments cannot overlap. But there is potentially a gap at the beginning and another at the end, that is, if `offsets[0] > 0` or if `offsets[N+1] > length(keys)`.  

Do you foresee any case where you would want to allow gaps between segments?  What about segments out of order?  (I think the non-overlapping segments is certainly a sanity-saver.)  Would the potential for gaps at the beginning or the end cause any problem?

adayton1: I believe both gaps and out of order segments will already work, but I will add test cases to verify. Definitely don't want to support overlapping segments.

test/TestSegmentedSort.cpp
Arlie-Capps, line 1: Consider adding tests for bogus segment input---i.e. non-monotonic offsets, or offsets that go before or beyond the end of the range.

And, consider warning users in the documentation that bad things will happen for bad segment arrays.

src/care/cuda/sort.h
adayton1: Oh, never mind, it's not possible to support them with this API. If we did want to support gaps or out of order segments, we could add another overload that takes one array for the segment starts and another array for the segment ends. I don't have a use case in mind, so I'm hesitant to add that overload now.

test/TestSegmentedSort.cpp
```diff
@@ -12,16 +12,22 @@
...
+      7, 2
+   };
    const int segmentOffsets[] = {0, 3, 3, 6, 8};
-   const int expected[] = {1, 4, 5, 3, 8, 9, 2, 7};
+   const int expected[] = {
+      1, 4, 5,
+      // empty segment
+      3, 8, 9,
+      2, 7
+   };
```
Arlie-Capps, line 30: Do you need to suppress the clang-formatter for this, or does Care not have the formatter turned on?

test/TestSegmentedSort.cpp
Arlie-Capps, line 40: hah.  I missed this line the first and second time through this routine.  I was very confused that this test didn't do anything.  Then I looked again and saw that the test does call `segmented_sort()`.

test/TestSegmentedSort.cpp
adayton1: The formatter is not turned on right now.

src/care/cuda/sort.h
adayton1: But I have clarified the documentation indicating that the offsets must go from 0 to keys.size() and be non-decreasing.

Arlie-Capps (approved):
I see you resisted the temptation to threaten the user with terrible doom and untraceable bugs if they give you a bogus offsets array... that restraint does fit with your style. : )

test/TestSegmentedSort.cpp
adayton1: I've updated the documentation. Adding checks in the algorithms would be pretty inefficient, at least for CUDA/HIP.

